PR #11, opened 2024-01-16
Share k8s worker and control-plane charms

addyess:
## Structure of the charms

The `k8s` and `k8s-worker` charms are noticeably tucked into one-another.  

```
└── worker
    ├── charmcraft.yaml
    ├── requirements.txt
    └── k8s
        ├── charmcraft.yaml
        ├── lib
        │   └── charms/...
        ├── requirements.txt
        └── src
            └── charm.py
```

While unfamiliar to some charm developers, this lets both charms share the exact same `src` folder. This is accomplished by using the `parts.charm.charm-entrypoint` value in the `worker` directory set to `k8s/src/charm.py`.

### What's unique

The unique parts of the charm are what are in each charm's top-level directory:

```
charmcraft.yaml
config.yaml
actions.yaml
metadata.yaml
requirements.yaml
```

In order to exclude the `k8s` exclusive components from the `k8s-worker` charm, charmcraft will read the `worker/.jujuignore` file to determine what to leave out of the final charm.

### What's not

The shared portions of each charm are within `worker/k8s` (except for the above mentioned exclusions).  This includes shared libraries from `worker/k8s/lib`, shared source from `worker/k8s/src`, shared python dependencies from `worker/k8s/requirements.txt`

### How to distinguish which charm code should engage

The charm can distinguish whether it's a `control-plane` or `worker` unit by using `self.is_worker` or `self.is_control_plane` by querying its metadata.

### Why two charms?

Much of the charm's behavior will be identical. They will employ many of the same relations, many of the same resources, configure the same snap, and use many of the same configuration options. One might therefore assume the two should be 1 charm. History with Charmed Kubernetes has proven that having 2 charms split between control-plane and worker has advantages when a relation is split across `requires` and `provides`.

### Why not use a charm library?

Sharing code between a charm library is a really reasonable idea, there are limitations that a charm library presents:
* limited to a single file
* PRs where the library changes doesn't reflect in the secondary charm
* updating a second charm isn't immediate 
    - must upload to charmhub, then download into the secondary charms

### How to use two charms in the same code base:

In cases where the charms should diverge the behavior, use a runtime switch to make the decision

```python
if self.is_control_plane:
    # do control-plane only thing
    ...
# do more common things
...
if self.is_worker:
    # do worker only thing
    ...
```


bschimke95:
What is the status here Adam?

addyess:
> What is the status here Adam?

I think i wanna get back into it.  Lemme resolve and see if we can get this guy rolling

addyess:
Really need the following PRs to land first
* https://github.com/canonical/operator-workflows/pull/248
* https://github.com/canonical/operator-workflows/pull/247

.github/workflows/auto-update-libs-k8s-worker.yaml
```diff
@@ -9,4 +9,4 @@ jobs:
     uses: canonical/operator-workflows/.github/workflows/auto_update_charm_libs.yaml@main
     secrets: inherit
     with:
-      working-directory:  ./charms/k8s-worker/
\ No newline at end of file
+      working-directory:  ./charms/k8s/worker/
```
bschimke95, line 12: nit: newline

bschimke95 (approved):
LGTM, only a minor comment

charms/k8s/worker/src/charm.py
```diff
@@ -180,7 +199,7 @@ def _reconcile(self, _):
         """Reconcile state change events."""
         self._install_k8s_snap()
         self._apply_snap_requirements()
-        if self.unit.is_leader():
+        if self.unit.is_leader() and self.is_control_plane:
```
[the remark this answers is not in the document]
addyess: every juju application gets a leader chosen for it by the juju controller.  Juju enforces only one unit in an application can be the leader.  `k8s` is one application and `k8s-worker` is another.  So -- each will have a selected leader. 

leader isn't intended to make decisions about the underlying application -- it should just be used as a convenience for having "one unit in the app to make a decision for all apps"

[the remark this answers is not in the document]
bschimke95: I see, thanks for the explanation!

charms/k8s/charmcraft.yaml
```diff
@@ -21,3 +59,20 @@ config:
...
     build-packages: [git]
+    charm-entrypoint: worker/src/charm.py
+  lib:
+    # move the ./worker/lib path to ./lib since 
+    # charmcraft assumes it to be there once the charm runs
+    after: [charm]
+    plugin: nil
+    source: ./
+    override-prime: |
+      mv worker/lib lib
```
addyess, line 70: So, it seems that charmcraft expects the `lib` folder to always been at the same path level as charmcraft.yaml and other "charm files" when run on the unit.  This moves the lib file to the top level during the charm build process.

charms/k8s/charmcraft.yaml
```diff
@@ -21,3 +59,20 @@ config:
...
+
+peers:
+  cluster:
+    interface: cluster
+
+provides:
+  tokens:
+    interface: tokens
```
addyess, line 78: @mateoflorido if this isn't the `tokens` interface?  What is it.  So far as i can tell on the worker unit the relation name should be `cluster` but the interface it self can be anything.  

ideas for the interface name (not relation name)
* k8s-cluster -- this is how to cluster other components via the `k8s add-node/join-node` syntaxt
* k8s-worker -- this is a special relation to only `k8s-worker` nodes
* kube-control -- use the existing kube-control from charmed-kubernetes
    -- this would give a false impression we could use the existing "kubernetes-worker" charm with this charm

mateoflorido: It's the same tokens interface. We created tokens in Charmed Kubernetes for sharing the k8s authentication tokens over relation data (as we couldn't use Juju secrets). I like `k8s-cluster`. Why do we want to include `kube-control`? If I recall correctly, we need it for the cloud providers and maybe the new k8s-e2e charm. Should we create a new interface to avoid confusion and prevent people from trying to integrate incompatible charms?

addyess: great. let's go with `k8s-cluster` for the interface name.  I'll change it here for now instead of being "tokens".  But won't provide an implementation for that relation

charms/k8s/charmcraft.yaml
```diff
@@ -21,3 +59,21 @@ config:
...
+    charm-entrypoint: worker/src/charm.py
+  lib:
+    # move the ./worker/lib path to ./lib since 
+    # charmcraft assumes it to be there once the charm runs
+    after: [charm]
+    plugin: nil
+    source: ./
+    override-prime: |
+      rm -rf $CRAFT_PRIME/lib
+      mv $CRAFT_PRIME/worker/lib $CRAFT_PRIME/lib
```
addyess, line 71: So, it seems that charmcraft expects the lib folder to always been at the same path level as charmcraft.yaml and other "charm files" when run on the unit. 

This moves the lib/ folder to the top level during the charm build process.

charms/k8s/worker/lib/charms/k8s/v0/k8sd_api_manager.py
mateoflorido, line 1: Is it possible to move the libraries to the top level directory? I think we can avoid running the prime in the charmcraft.yaml, and it makes more sense to store the library in the charm that actually owns it.

mateoflorido (commented):
Just a brief comment

PR #12, opened 2024-01-26
Bootstrap k8s-operator

mateoflorido:
## Overview

Generate a first version of the `k8s-operator` charm.

### Rationale

This pull request introduces an initial version of the `k8s-operator`, which has the option to bootstrap and deploy a one-node cluster. This is a stepping stone for future work that has to be done in the charm.

### Juju Events Changes

- Refactored `on_install` handler
- Refactored `update_status` handler

### Library Changes
- Bumped `k8s_api_manager` patch.
- Added Pydantic models and validators for the ClusterStatus endpoint.
- Added methods for checking k8sd cluster state

### DIFF
--- a/charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
+++ b/charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
@@ -43,7 +43,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 1
 
 logger = logging.getLogger(__name__)
 
@@ -144,10 +144,10 @@ class CreateJoinTokenResponse(BaseRequestModel):
     """Response model for join token creation requests.
 
     Attributes:
-        metadata (TokenMetadata): Metadata containing the newly created join token.
+        token (str): Metadata containing the newly created join token.
     """
 
-    metadata: TokenMetadata
+    token: str = Field(..., alias="metadata")
 
 
 class ClusterMember(BaseModel):
@@ -213,31 +213,6 @@ class GetClusterStatusResponse(BaseRequestModel):
     """
 
     metadata: Optional[ClusterMetadata] = None
-
-    @validator("error_code", always=True)
-    def check_error_code(cls, v, values):
-        """Override the base model validator for error_code to allow any value.
-
-        Args:
-            v (int): The value of the error_code field.
-            values (dict): The values dictionary.
-
-        Returns:
-            int: The unvalidated error code.
-        """
-        return v
-
-    @validator("status_code", always=True)
-    def check_status_code(cls, v):
-        """Override the base model validator for status_code to allow any value.
-
-        Args:
-            v (int): The value of the status_code field.
-
-        Returns:
-            int: The unvalidated status code.
-        """
-        return v
 
 
 T = TypeVar("T", bound=BaseRequestModel)
@@ -373,17 +348,22 @@ class K8sdAPIManager:
         """
         try:
             with self.factory.create_connection() as connection:
-                if body is not None:
-                    body_data = json.dumps(body)
-                else:
-                    body_data = None
+                body_data = json.dumps(body) if body is not None else None
+                headers = {"Content-Type": "application/json"} if body_data else {}
+
                 connection.request(
                     method,
                     endpoint,
                     body=body_data,
-                    headers={"Content-Type": "application/json"},
+                    headers=headers,
                 )
                 response = connection.getresponse()
+
+                if not 200 <= response.status < 300:
+                    raise InvalidResponseError(
+                        f"Request failed with status {response.status}: {response.reason}"
+                    )
+
                 data = response.read().decode()
             return response_cls.parse_raw(data)
 
@@ -401,10 +381,10 @@ class K8sdAPIManager:
         Returns:
             str: The generated join token if successful.
         """
-        endpoint = "/1.0/k8sd/tokens"
+        endpoint = "/cluster/1.0/tokens"
         body = {"name": name}
         join_response = self._send_request(endpoint, "POST", CreateJoinTokenResponse, body)
-        return join_response.metadata.token
+        return join_response.token
 
     def enable_component(self, name: str, enable: bool):
         """Enable or disable a k8s component.
@@ -423,9 +403,13 @@ class K8sdAPIManager:
         Returns:
             bool: True if the cluster has been bootstrapped, False otherwise.
         """
-        endpoint = "/1.0/k8sd/cluster"
-        cluster_status = self._send_request(endpoint, "GET", GetClusterStatusResponse)
-        return cluster_status.error_code == 0
+        try:
+            endpoint = "/1.0/k8sd/cluster"
+            cluster_status = self._send_request(endpoint, "GET", GetClusterStatusResponse)
+            return cluster_status.error_code == 0
+        except InvalidResponseError as e:
+            logger.error("Invalid response while checking if cluster is bootstrapped: %s", e)
+        return False
 
     def is_cluster_ready(self):
         """Check if the Kubernetes cluster is ready.
--- a/charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
+++ b/charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
@@ -27,10 +27,11 @@ Similarly, the module allows for requesting authentication tokens and
 managing K8s components.
 """
 import json
+import logging
 import socket
 from contextlib import contextmanager
 from http.client import HTTPConnection, HTTPException
-from typing import List, Type, TypeVar
+from typing import List, Optional, Type, TypeVar
 
 from pydantic import BaseModel, Field, validator
 
@@ -42,7 +43,9 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
+
+logger = logging.getLogger(__name__)
 
 
 class K8sdAPIManagerError(Exception):
@@ -94,11 +97,12 @@ class BaseRequestModel(BaseModel):
         return v
 
     @validator("error_code", always=True)
-    def check_error_code(cls, v):
+    def check_error_code(cls, v, values):
         """Validate the error_code field.
 
         Args:
             v (int): The value of the error_code field to validate.
+            values (dict): The values dictionary.
 
         Returns:
             int: The validated error code if it is 0.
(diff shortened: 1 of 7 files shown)


charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -145,6 +150,96 @@ class CreateJoinTokenResponse(BaseRequestModel):
...
+
+class GetClusterStatusResponse(BaseRequestModel):
+    """Response model for getting the status of the k8sd cluster.
+
+    Attributes:
+        metadata (Optional[ClusterMetadata]): Metadata containing the cluster status.
+                                              Can be None if the status is not available.
+    """
+
+    metadata: Optional[ClusterMetadata] = None
```
bschimke95, line 215: Why do you need to wrap this here?

The Go type does not have a `metadata` member:
```
// GetClusterStatusResponse is the response for "GET 1.0/k8sd/cluster".
type GetClusterStatusResponse struct {
	ClusterStatus ClusterStatus `json:"status"`
}
```

charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -276,10 +373,14 @@ def _send_request(self, endpoint: str, method: str, body: dict, response_cls: Ty
         """
         try:
             with self.factory.create_connection() as connection:
+                if body is not None:
+                    body_data = json.dumps(body)
```
bschimke95, line 377: You could do:
```
body_data = body if body is not None else json.dumps(body)
```
Up to you if this improves readability or not :)

charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -302,7 +403,7 @@ def create_join_token(self, name: str):
         """
         endpoint = "/1.0/k8sd/tokens"
```
bschimke95, line 404: The tokens endpoint was removed in https://github.com/canonical/k8s-snap/pull/73
The additional logic around the token generation moved into the microcluster hooks which made this endpoint obsolete. 

Instead, directly query the microcluster endpoint `cluster/1.0/tokens` to issue a new token.
See: 
https://github.com/canonical/microcluster/blob/1525f8ea8d7a2196d7825bc3ca5fc29368a0fd1b/internal/rest/client/tokens.go#L18
and the type definition here:
https://github.com/canonical/microcluster/blob/1525f8ea8d7a2196d7825bc3ca5fc29368a0fd1b/internal/rest/types/tokens.go#L11

charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -313,8 +414,30 @@ def enable_component(self, name: str, enable: bool):
...
+
+    def is_cluster_bootstrapped(self) -> bool:
+        """Check if K8sd has been bootstrapped.
+
+        Returns:
+            bool: True if the cluster has been bootstrapped, False otherwise.
+        """
+        endpoint = "/1.0/k8sd/cluster"
+        cluster_status = self._send_request(endpoint, "GET", GetClusterStatusResponse)
+        return cluster_status.error_code == 0
```
bschimke95, line 428: I wonder why we have this extra layer of error codes? Can we not use the HTTP response code?

charms/k8s/src/charm.py
```diff
@@ -16,35 +16,143 @@
...
+        commands = [
+            "/snap/k8s/current/k8s/connect-interfaces.sh",
+            "/snap/k8s/current/k8s/network-requirements.sh",
+        ]
+        for c in commands:
+            subprocess.check_call(shlex.split(c))
+
+    def _bootstrap_k8s_snap(self):
+        """Bootstrap the k8s if it's not already bootstrapped."""
+        if not self.api_manager.is_cluster_bootstrapped():
```
bschimke95, line 75: @neoaggelos I guess calling `k8s bootstrap` on a bootstrapped cluster should just noop and succeed, WDYT?

bschimke95 (commented):
Great work, Mateo!
Left some comments.

charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -145,6 +150,96 @@ class CreateJoinTokenResponse(BaseRequestModel):
     metadata: TokenMetadata
 
 
+class ClusterMember(BaseModel):
```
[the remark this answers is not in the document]
neoaggelos: i'd expect that bringing in protocol buffers here would be an even bigger hustle

charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
neoaggelos: microclusters puts the response in a `metadata` field, along with others to manage errors etc.

charms/k8s/src/charm.py
neoaggelos: no, bootstrap is not an idempotent action

charms/k8s/src/charm.py
```diff
@@ -16,35 +16,143 @@
...
+
+        Args:
+            event: The event that triggered this handler.
+        """
+        try:
+            self._install_k8s_snap()
+            self._apply_snap_requirements()
+            self._bootstrap_k8s_snap()
+            self._enable_components()
+            self.unit.status = WaitingStatus("K8s not ready")
```
ktsakalozos, line 123: This line is a bit confusing. Why isn't this ops.WaitingStatus ? And why does it way that k8s is not ready? Are we waiting for the update hook to run to get the status updated to ready? If yes, then that would be after how long?

charms/k8s/src/charm.py
```diff
@@ -16,35 +16,143 @@
...
+    def _on_install(self, event):
+        """Handle install event for the charm.
+
+        Args:
+            event: The event that triggered this handler.
+        """
+        try:
+            self._install_k8s_snap()
+            self._apply_snap_requirements()
+            self._bootstrap_k8s_snap()
```
ktsakalozos, line 121: I think only the leader of the charm units should bootstrap and the rest should just join. Right @bschimke95 ?

charms/k8s/src/charm.py
```diff
@@ -16,35 +16,143 @@
...
+
+    def _apply_snap_requirements(self):
+        """Apply necessary snap requirements for the k8s snap.
+
+        This method executes necessary scripts to ensure that the snap
+        meets the network and interface requirements.
+        """
+        self.unit.status = ops.MaintenanceStatus("Applying K8s requirements")
+        commands = [
+            "/snap/k8s/current/k8s/connect-interfaces.sh",
```
ktsakalozos, line 67: This is only needed when we deploy a local built of the snap.

charms/k8s/src/charm.py
bschimke95: yes, but I think Mateo only targets single node in this PR. Clustering will probably be handled in another PR.
@mateoflorido You can add a TODO to remember for the follow up, but IMHO this is fine for now.

charms/k8s/src/charm.py
bschimke95: Hm I agree that this is not idempotent, thatswhy I wrote it is a noop. 

But as a user I want to bootstrap a cluster - if the cluster is already bootstrapped, that's great -> the command should succeed because my intent was fullfilled. 

We shouldn't offload that check to the caller, he doesn't know any internals and then needs to guess if it is bootstrapped already (e.g. calling some endpoint like Mateo did). 
However, this call can fail for various reasons and not just because of an unitialized cluster.

charms/k8s/src/charm.py
```diff
@@ -16,35 +16,143 @@
...
+        ]
+        for c in commands:
+            subprocess.check_call(shlex.split(c))
+
+    def _bootstrap_k8s_snap(self):
+        """Bootstrap the k8s if it's not already bootstrapped."""
+        if not self.api_manager.is_cluster_bootstrapped():
+            self.unit.status = ops.MaintenanceStatus("Bootstrapping Cluster")
+            cmd = "k8s bootstrap"
+            subprocess.check_call(shlex.split(cmd), shell=False)
```
ktsakalozos, line 78: I thought we had an endpoint to do this.

bschimke95: We are not there yet.
For the JSON RPC endpoints to work, Microcluster expects to be initialized first. 
So, right now we have some logic in the CLI to do that.
https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8s/client/cluster.go#L30

We need to address this, but since there is an OK workaround for this, I guess it is not too critical right now.

charms/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -42,7 +43,9 @@ class utilises different connection factories (UnixSocketConnectionFactory
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
```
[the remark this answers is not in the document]
mateoflorido: You're right. The charm is [there](https://charmhub.io/k8s), but we are not publishing the library. Let me rollback the patch increment.

charms/k8s/src/charm.py
mateoflorido: I'll refactor this fragment. We should directly invoke a set_status handler and finish the event. If it's not ready, we will wait until the update_status handler is run. Regarding the periodicity of the update_status events, they should be fired at 5 min intervals; however, users can change this value.

charms/k8s/src/charm.py
mateoflorido: Correct, this has to be implemented in another PR. At the moment, we cannot add TODOs because the CI will complain that there are TODOs in the code. I'll update the CI suite to address this issue.

bschimke95: WTF? I understand that automatic code checking etc. can be great but it feels like you guys fight that thing more often then as it is useful...

addyess (changes requested):
Just to leave my two cents, i love the work in the library and tests.  All excellent

I really don't care for the  `event.defer()`  showing up for every event the charm responds to.  that's not gonna work out.  I'd prefer something that has no `event.defer()` like the reconciler pattern

charms/k8s/src/charm.py
```diff
@@ -16,35 +16,143 @@
...
+
+        logger.info("Snap k8s not found or no version available.")
+        return None
+
+    def _install_k8s_snap(self):
+        """Install the k8s snap package."""
+        self.unit.status = ops.MaintenanceStatus("Installing k8s snap")
+        k8s_snap = self.snap_cache["k8s"]
+        if not k8s_snap.present:
+            k8s_snap.ensure(SnapState.Latest, channel="edge")
```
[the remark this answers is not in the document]
mateoflorido: I'm going to create the channel configuration. In the future, we can plan a process to determine how to automatically figure out the charm channel.

charms/k8s/src/charm.py
mateoflorido: It sounds like we are relying on `connect-interfaces.sh` to enable some interfaces in the snap. If I remove the call, the charm does not become active.

addyess (approved):
yeah... really nice!

PR #13, opened 2024-01-29
Implement  `reconcile` pattern

mateoflorido:
## Overview
Refactor the charm's business logic to follow the reconcile pattern.
### Rationale
Due to the unstable and unpredictable behavior resulting from using `event.defer()` in the charm, it is recommended to fallback to a more robust mechanism. This pull request introduces a refactor in the charm's business logic, aligning it with the reconcile pattern. This change involves the usage of two libraries: [charm-lib-contextual-status](https://github.com/charmed-kubernetes/charm-lib-contextual-status) and [charm-lib-reconciler](https://github.com/charmed-kubernetes/charm-lib-reconciler).
### Juju Events Changes
- Modified `update-status` handler.
- Added `reconciler` handler.

### DIFF
--- a/charms/k8s/requirements.txt
+++ b/charms/k8s/requirements.txt
@@ -1,4 +1,4 @@
 ops >= 2.2.0
 pydantic == 1.*
 charm-lib-reconciler @ git+https://github.com/charmed-kubernetes/charm-lib-reconciler
-charm-lib-contextual-status @ git+https://github.com/charmed-kubernetes/charm-lib-contextual-status@feat/custom-exit-status
+charm-lib-contextual-status @ git+https://github.com/charmed-kubernetes/charm-lib-contextual-status
--- a/charms/k8s/requirements.txt
+++ b/charms/k8s/requirements.txt
@@ -1,2 +1,4 @@
 ops >= 2.2.0
 pydantic == 1.*
+charm-lib-reconciler @ git+https://github.com/charmed-kubernetes/charm-lib-reconciler
+charm-lib-contextual-status @ git+https://github.com/charmed-kubernetes/charm-lib-contextual-status@feat/custom-exit-status
--- a/charms/k8s/src/charm.py
+++ b/charms/k8s/src/charm.py
@@ -23,6 +23,7 @@ from typing import Optional
 
 import charms.contextual_status as status
 import ops
+from charms.contextual_status import WaitingStatus, on_error
 from charms.k8s.v0.k8sd_api_manager import (
     InvalidResponseError,
     K8sdAPIManager,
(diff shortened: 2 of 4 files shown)


charms/k8s/requirements.txt
```diff
@@ -1,2 +1,4 @@
 ops >= 2.2.0
 pydantic == 1.*
+charm-lib-reconciler @ git+https://github.com/charmed-kubernetes/charm-lib-reconciler
+charm-lib-contextual-status @ git+https://github.com/charmed-kubernetes/charm-lib-contextual-status@feat/custom-exit-status
```
addyess, line 4: should we get this merged first?

charms/k8s/src/charm.py
```diff
@@ -108,53 +130,25 @@ def _install_k8s_snap(self):
...
                 if version := self._get_snap_version():
                     self.unit.set_workload_version(version)
-                self.unit.status = ops.ActiveStatus("Ready")
+            else:
+                status.add(ops.WaitingStatus("Waiting for k8s to be ready."))
         except InvalidResponseError:
-            self.unit.status = ops.WaitingStatus("Waiting for K8s to be ready")
-            event.defer()
+            log.info("Invalid Response from K8sd...")
+            status.add(ops.WaitingStatus("Waiting for K8s to be ready"))
```
bschimke95, line 143: This will stay in the `WaitingStatus` forever if there is a bug in `k8sd`. Should we change into `ErrorStatus` after some retries?

bschimke95 (commented):
LGTM, minor comment.
Leaving the approval to the charming expert Adam.

charms/k8s/src/charm.py
mateoflorido: Correct. Personally, I believe the charm must be agnostic (at certain level) to the underlying machinery (i.e., k8s-snap). Our CI machinery only passes the build/deploy stage if the charm becomes active/idle. Handling this retry might involve implementing a busy-waiting mechanism or storing the state, which is not the best option in certain cases. Could we improve the snap side to expose a mechanism that flags when the snap is unable to come up?

charms/k8s/requirements.txt
mateoflorido: I think we need to explore the actual improvements we want to implement in the libraries. So, I'll point to the main branches.

addyess (approved):
All of this is so much better than the defers()

I like the inclusion of the `_update_status()` at the end of the reconcile loop alot!

charms/k8s/src/charm.py
```diff
@@ -125,31 +119,30 @@ def _get_snap_version(self) -> Optional[str]:
...
-                status.add(ops.WaitingStatus("Waiting for k8s to be ready."))
-        except InvalidResponseError:
-            log.info("Invalid Response from K8sd...")
-            status.add(ops.WaitingStatus("Waiting for K8s to be ready"))
-        except K8sdConnectionError:
-            log.exception("Unable to contact K8sdAPI")
-            status.add(ops.WaitingStatus("Waiting for K8sd API"))
+        if self.api_manager.is_cluster_ready():
+            if version := self._get_snap_version():
+                self.unit.set_workload_version(version)
```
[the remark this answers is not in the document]
mateoflorido: Nice catch, fixed!

charms/k8s/src/charm.py
```diff
@@ -97,64 +116,38 @@ def _get_snap_version(self) -> Optional[str]:
...
-                self.unit.status = ops.ActiveStatus("Ready")
-        except InvalidResponseError:
-            self.unit.status = ops.WaitingStatus("Waiting for K8s to be ready")
-            event.defer()
-        except K8sdConnectionError:
-            logger.exception("Unable to contact K8sdAPI")
-            self.unit.status = ops.WaitingStatus("Waiting for K8sd API")
-            event.defer()
+    @on_error(
+        ops.WaitingStatus("Failed to update status"),
```
addyess, line 132: maybe something like this -- we're "waiting" because it could be ready later right?  And the user doesn't need to do anything?

```suggestion
        ops.WaitingStatus("Cluster not yet ready"),
```

addyess (approved):
LGTM 🚀 🍰

PR #14, opened 2024-02-01
Use API instead of CLI for bootstrapping

bschimke95:
### Overview

Directly call the `k8s-snap` API endpoints instead of the CLI.

### Rationale

The k8s-snap should directly interact with the API instead of using the CLI as a proxy for those requests.

### Library Changes

* add `check_k8sd_ready` and `bootstrap_k8s_snap`
* rename `UpdateComponentResponse` to `EmptyResponse` to make it reusable
* fix `None` check in JSON body

### DIFF
--- a/charms/k8s/src/charm.py
+++ b/charms/k8s/src/charm.py
@@ -64,6 +64,7 @@ class K8sCharm(ops.CharmBase):
         # TODO: Implement clustering using leader units.
         self._install_k8s_snap()
         self._apply_snap_requirements()
+        self._check_k8sd_ready()
         self._bootstrap_k8s_snap()
         self._enable_components()
         self._update_status()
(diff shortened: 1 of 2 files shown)


charms/k8s/src/charm.py
```diff
@@ -83,13 +84,30 @@ def _apply_snap_requirements(self):
...
             status.add(ops.MaintenanceStatus("Bootstrapping Cluster"))
-            cmd = "k8s bootstrap"
-            subprocess.check_call(shlex.split(cmd))
+            binding = self.model.get_binding("juju-info")
+            address = binding and binding.network.ingress_address
+            # k8s/x to k8s-x to avoid trouble with urls
+            name = self.unit.name.replace("/", "-")
+
+            # TODO: Make port (and address) configurable.
+            self.api_manager.bootstrap_k8s_snap(name, f"{str(address)}:6400")
```
addyess, line 110: @mateoflorido shouldn't this bootstrapping really only occur on the charm leader?  It seems to me we can't bootstrap every control-plane node independently

bschimke95: Clustering is not part of this PR. Let's add this check once we need it.

addyess: 👍🏻 agree, i see that on the clustering PR.  No worries

addyess (approved):
Nice work

PR #15, opened 2024-02-02
Implement `k8s` clustering

mateoflorido:
## Overview

Implement the Join Cluster process for `k8s` charm units.

### Rationale

This pull request introduces the implementation for joining multiple control plane units into the cluster via the `cluster` peer relation. This relation utilizes Juju secrets for distributing the cluster join tokens over the relation data.

### Juju Events Changes

- Handlers have been added for the cluster relation (`create_cluster_tokens` and `join_cluster`).

### DIFF
--- a/charms/k8s/src/charm.py
+++ b/charms/k8s/src/charm.py
@@ -43,8 +43,6 @@ K8SD_SNAP_SOCKET = "/var/snap/k8s/common/var/lib/k8sd/state/control.socket"
 class K8sCharm(ops.CharmBase):
     """A charm for managing a K8s cluster via the k8s snap."""
 
-    _stored = ops.StoredState()
-
     def __init__(self, *args):
         """Initialise the K8s charm.
 
@@ -60,7 +58,6 @@ class K8sCharm(ops.CharmBase):
         self.reconciler = Reconciler(self, self._reconcile)
 
         self.framework.observe(self.on.update_status, self._on_update_status)
-        self._stored.set_default(joined=False)
 
     @on_error(WaitingStatus("Failed to apply snap requirements"), subprocess.CalledProcessError)
     def _apply_snap_requirements(self):
@@ -165,7 +162,7 @@ class K8sCharm(ops.CharmBase):
     @on_error(WaitingStatus("Waiting for Cluster token"), TypeError)
     def _join_cluster(self):
         """Retrieve the join token from secret databag and join the cluster."""
-        if self.unit.is_leader() or self._stored.joined:
+        if self.api_manager.is_cluster_bootstrapped():
             return
 
         status.add(ops.MaintenanceStatus("Joining cluster"))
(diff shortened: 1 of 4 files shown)


charms/k8s/src/charm.py
```diff
@@ -128,6 +145,36 @@ def _install_k8s_snap(self):
...
+    def _reconcile(self, _):
+        """Reconcile state change events."""
+        self._install_k8s_snap()
+        self._apply_snap_requirements()
+        if self.unit.is_leader():
+            self._bootstrap_k8s_snap()
+            self._enable_components()
+            self._create_cluster_tokens()
+        else:
+            self._join_cluster()
```
addyess, line 175: ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

charms/k8s/src/charm.py
```diff
@@ -128,6 +145,36 @@ def _install_k8s_snap(self):
...
+            cmd = f"k8s join-cluster {shlex.quote(token)}"
+            subprocess.check_call(shlex.split(cmd))
+            self._stored.joined = True
+
+    def _reconcile(self, _):
+        """Reconcile state change events."""
+        self._install_k8s_snap()
+        self._apply_snap_requirements()
+        if self.unit.is_leader():
+            self._bootstrap_k8s_snap()
```
addyess, line 171: what if `_bootstrap_k8s_snap()` sets the `self._stored.joined=True`??  I don't suppose there's anyway to ask the snap itself if this node is part of the cluster?

mateoflorido: Yes, I have a better idea to avoid using StoredState, we can leverage the `is_cluster_bootstrapped check. Thanks for starting my train of thought @addyess 🚆

charms/k8s/src/charm.py
```diff
@@ -146,6 +159,35 @@ def _install_k8s_snap(self):
             channel = self.config["channel"]
             k8s_snap.ensure(SnapState.Latest, channel=channel)
 
+    @on_error(WaitingStatus("Waiting for Cluster token"), TypeError)
+    def _join_cluster(self):
+        """Retrieve the join token from secret databag and join the cluster."""
+        if self.api_manager.is_cluster_bootstrapped():
```
addyess, line 165: ❤️

addyess (commented):
This is looking so nice... a few questions still

charms/k8s/src/charm.py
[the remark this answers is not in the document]
addyess: I agree with you that the stored state removal is a great improvement

Is the
```python
else:
    self._join_cluster()
```

necessary or can it be 
```python
self._join_cluster()
```

would the leader control-plane unit be okay with this call because it would run into 

```python
if self.api_manager.is_cluster_bootstrapped():
    return
```

and return?

addyess (approved):
100% LGTM

PR #20, opened 2024-02-09
Implement node removal

bschimke95:
### Overview

Cleanup cluster if a unit departs from the cluster. Specifically it will,
1. Cleanup the token secret from the `cluster` relation 
2. Remove the node from the cluster by triggering the cluster removal Endpoint of the `k8s snap`

### Module Changes

* Add `remove_node` function that is executed only on the leader control-plane node.

### Library Changes

* Add `remove_node` API call


charms/worker/k8s/src/charm.py
```diff
@@ -200,6 +200,40 @@ def _join_cluster(self):
...
+
+            status.add(ops.MaintenanceStatus("Removing nodes"))
+
+            for unit in units_to_remove:
+                del app_databag[self.unit.name]
+
+                node_name =unit.replace("/", "-")
+                # TODO: This will currently fail because worker removal does not work yet.
+                self.api_manager.remove_node(node_name)
+
```
mateoflorido, line 236: What if we leverage set operations instead?
```suggestion
    @on_error(WaitingStatus("Waiting to remove nodes"), TypeError)
    def _remove_nodes(self):
        """Check for units that where removed, revoke their secrets and remove them from the k8s cluster."""
        if relation := self.model.get_relation("cluster"):
            units = {u.name for u in relation.units}
            app_databag = relation.data.get(self.model.app, {})

            units_to_remove = set(app_databag.keys()) - units
            
            if not units_to_remove:
                return
                
            status.add(ops.MaintenanceStatus("Removing nodes"))

            for unit_name in units_to_remove:
                log.info(f" Unit: {unit_name} leaving cluster.")
                del app_databag[unit_name]

                node_name = unit_name.replace("/", "-")
                # TODO: This will currently fail because worker removal does not work yet.
                self.api_manager.remove_node(node_name)
```

mateoflorido (changes requested):

charms/worker/k8s/src/charm.py
addyess: the [`ops.RelationDepartedEvent`](https://ops.readthedocs.io/en/latest/#ops.RelationDepartedEvent) will tell you which unit is leaving.  Would that help?

mateoflorido: That would help. But do we want that hybrid approach here? Some functions receiving the event, others don't...

addyess: meh -- it's just there if you want it.  If the event happens to be the right one on the right relation -- it could be useful

bschimke95: So, if we still need the set check I would just use that and not two different checks to archive the same thing. I understand that this would have a slightly better performance (we don't need to search through all units/secrets if we can facilitate the `RelationDeparted` event) but I would rather go for readability at this point. WDYT?

tests/integration/test_k8s.py
```diff
@@ -44,7 +45,19 @@ async def ready_nodes(k8s, expected_count):
...
+async def test_nodes_ready(kubernetes_cluster: model.Model):
     """Deploy the charm and wait for active/idle status."""
     k8s = kubernetes_cluster.applications["k8s"]
     await ready_nodes(k8s.units[0], 3)
+
+
+@pytest.mark.abort_on_fail
+async def test_remove_units(kubernetes_cluster: model.Model):
+    """Deploy the charm and wait for active/idle status."""
+    k8s: application.Application = kubernetes_cluster.applications["k8s"]
```
bschimke95, line 57: This will spin up a whole new k8s cluster, right? 
What is the best practice here - should we merge both tests or is do we accept the additional testing time?

tests/integration/test_k8s.py
```diff
@@ -44,7 +45,19 @@ async def ready_nodes(k8s, expected_count):
...
+    await ready_nodes(k8s.units[0], 3)
+
+    kubernetes_cluster.destroy_unit(k8s.units[1])
+    await ready_nodes(k8s.units[0], 2)
+    k8s.destroy_relation("k8s", "k8s-worker:cluster")
+    await ready_nodes(k8s.units[0], 1)
```
bschimke95, line 63: This will not work. 
The relation will be departed but ofc it takes some time for the event to be processed. 
Is there a clean way in Juju to wait for this?

tests/integration/test_k8s.py
addyess: nope, it should reuse the existing fixture since it has module scope, it will apply to all the tests in this python module

charms/worker/k8s/src/token_distributor.py
```diff
@@ -147,48 +160,153 @@ def _create_cos_token(self, name: str, _):
...
+
+        log.info(
+            "Revoke report for %s \n\tjoined=%s\n\tremoving=%s\n\tremaining=%s",
+            relation.name,
+            ",".join(sorted(u.name for u in joined)),
+            ",".join(sorted(u.name for u in remove)),
+            ",".join(sorted(u.name for u in remaining)),
+        )
+
+        token_strat = self.token_repeaing_strategies.get(token_strategy)
```
louiseschmidtgen, line 297: Little typo here: repeating

bschimke95 (commented):
Tested locally and reviewed (only some typo fixes and rebase).

@mateoflorido please approve so that we can merge this.

mateoflorido (approved):
LGTM!

PR #29, opened 2024-02-21
Add AlertManager rules

mateoflorido:
## Overview

Implement the AlertManager integration via `cos-agent` for `k8s` and `k8s-worker` charms.

## Rationale

This pull request introduces an implementation for sending alert rules to Prometheus via relation data.

This version of the integration relies on manual testing. Integration tests will be introduced in future PRs.

### Module Changes

- Introduced a new file in the `scripts` folder, which helps pull the alerts from the upstream repository. It also ensures that the alerts are in the correct format and applies patches to fix some alerts expressions.

### Other Changes
- Added a new environment for updating the dashboards via `tox`.


addyess (approved):
Also LGTM

PR #31, opened 2024-02-23
Chore: Crashdump as root with the journalctl from k8s snap

addyess:
Run integration tests with juju-crashdump so that CI failures can be evaluated


kwmonroe (approved):
noice

PR #32, opened 2024-02-23
Chore/reusable kubernetes fixture

addyess:
Improves the `kubernetes_cluster` fixture to make it more reusable

Key improvements:
* if a user runs without `--model=whatever`, a new model is generated -- a fresh deployment occurs
* if a user runs with `--model=whatever` and that model doesn't exist -- a fresh deployment occurs
* if a user runs with `--model=whatever` and that model exists, but has no apps -- a fresh deployment occurs
* if a user runs with `--model=whatever` and that model exists, with apps -- an upgraded deployment occurs
* It uses an honest-to-gosh bundle file to reason about the charms deployed and relations.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -5,10 +5,10 @@
 import asyncio
 import contextlib
 import logging
-from dataclasses import asdict, dataclass
+from dataclasses import asdict, dataclass, field
 from itertools import chain
 from pathlib import Path
-from typing import List, Optional
+from typing import Any, List, Mapping, Optional
 
 import pytest
 import pytest_asyncio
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -162,16 +167,9 @@ async def deploy_model(
...
         async with ops_test.fast_forward():
-            await asyncio.gather(
-                *(
-                    the_model.deploy(**asdict(charm))
-                    for charm in deploy_args
-                    if charm.application_name not in the_model.applications
-                )
-            )
-            await the_model.integrate("k8s", "k8s-worker:cluster")
+            await the_model.deploy(bundle.render)
```
addyess, line 170: pylibjuju knows how to deploy a bundle over an existing model and "do the right thing"

it will even upgrade any apps in the model with the fresh charms!! WOOT!

kwmonroe (approved):
LGTM

PR #34, opened 2024-02-26
the snaps set the node-role labels, the charm only now worries about custom labels

addyess:
### Overview

Resolves an issue where the charm didn't remove labels it had added via config.
Removes unnecessary management of the node role label

### Rationale

in order to use `ops.StoredState`, the `ops.Object` in `LabelMaker` must have a lifetime equal to the `ops.CharmBase` otherwise the framework cannot run it's `on_commit` handler to actually save the state. 

This PR  changes `LabelMaker(...)` to be owned by the Charm itself as an attribute (`self._labeler`) rather than constructing the labeler in the `apply_node_labels` handler


mateoflorido (approved):
LGTM!

PR #36, opened 2024-02-27
Improve Cluster Metrics Labeling

mateoflorido:
## Overview

Improve cluster metrics labeling
### Rationale

Many of the metrics were being scraped with labels that didn't allow Grafana to differentiate among instances. Therefore, the metrics were shown as a single unit, instead of rendering different instances. Below is an example of the raw metrics and how they were rendered.

Screenshot from 2024-02-27 14-43-26
Screenshot from 2024-02-27 14-43-46

After applying a new set of labels, the metrics are now shown correctly, each labeled with its respective host.

Screenshot from 2024-02-27 16-15-01
Screenshot from 2024-02-27 15-49-40

Finally, as Kube State Metrics is a global source of metrics, it does not require labeling with a set of unique metrics. static_configs were introduced as a way to override the default configurations to better define the metrics labels.

### DIFF
--- a/charms/worker/k8s/src/cos_integration.py
+++ b/charms/worker/k8s/src/cos_integration.py
@@ -5,7 +5,7 @@
 
 import logging
 from dataclasses import dataclass
-from typing import Dict, List, Union
+from typing import Any, Dict, List, Optional, Sequence, Union
 
 import ops
 from ops.charm import CharmBase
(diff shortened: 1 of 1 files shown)


bschimke95 (approved):
LGTM

Thanks for the great explanation in the PR description!

PR #38, opened 2024-03-01
Reduce charm lxd-profile to facilitate safe deployment

addyess:
Reduce the snap profile to the subset of options so that juju can safely deploy the charm without warnings/errors


mateoflorido (approved):
LGTM!

charms/worker/k8s/lxd-profile.yaml
```diff
@@ -12,19 +11,12 @@ config:
   security.nesting: "true"
   security.privileged: "true"
 devices:
-  aadisable:
-    path: /sys/module/nf_conntrack/parameters/hashsize
-    source: /sys/module/nf_conntrack/parameters/hashsize
-    type: disk
   aadisable2:
```
kwmonroe, line 14:
```suggestion
  aadisable:
```

charms/worker/lxd-profile.yaml
```diff
@@ -12,19 +11,12 @@ config:
   security.nesting: "true"
   security.privileged: "true"
 devices:
-  aadisable:
-    path: /sys/module/nf_conntrack/parameters/hashsize
-    source: /sys/module/nf_conntrack/parameters/hashsize
-    type: disk
   aadisable2:
```
kwmonroe, line 14:
```suggestion
  aadisable:
```

kwmonroe (approved):
lgtm with a couple tweaks.  those `aadisable` keys monotonically increase, so if you dropped the un-suffixed one, may as well `s/2//`.

neoaggelos (approved):
Note that `/dev/loop-control` is there but no loop devices.

If someone wants to use the default storage, they would have to manually set up loop devices in their profile

addyess:
> Note that `/dev/loop-control` is there but no loop devices.
> 
> If someone wants to use the default storage, they would have to manually set up loop devices in their profile

alrighty -- how many dev loops should there be?  The e2e tests make 16.  For lxd deployments -- what's a reasonable number of PVs to use?  When i deploy with juju -- i'm likely going to use cloud storage resources and not local -- but not really in the case of LXD. I don't have much experience with deploying to lxd since charmed-k8s didn't do that often.

addyess:
Actually @neoaggelos  -- back to this -- why can't we use loop storage from juju?

https://juju.is/docs/juju/storage-provider#heading--loop-devices-and-lxd

neoaggelos:
> Actually @neoaggelos -- back to this -- why can't we use loop storage from juju?

> https://juju.is/docs/juju/storage-provider#heading--loop-devices-and-lxd

So, to use loop devices from the LXD container, the respective `/dev/loop$x` must be in the profile, as shown in the docs you mentioned. It is true, we have 16 devices as a starting point in the snap, because that's enough for the e2e tests in our CI.

A container can have multiple profiles, so it does not make a difference if they come from the charm, or manual user config (as shown in the docs above).

As long as this is all clearly documented, then it's fine to not include them in the default profile.

As for "how many is enough", there is no right answer. There are `255` loop devices available in total (mapping to the respective `minor` numbers on the devices. It all depends. So, perhaps documentation is the way to go here.

eaudetcobello:
ubuntu@juju-a89ec2-1:~$ sudo k8s kubectl get pods -A
NAMESPACE     NAME                               READY   STATUS    RESTARTS   AGE
kube-system   cilium-kbqgg                       1/1     Running   0          3m51s
kube-system   cilium-l9rk4                       1/1     Running   0          105s
kube-system   cilium-operator-597b55b886-9ptf4   1/1     Running   0          3m51s
kube-system   cilium-rjnqn                       1/1     Running   0          2m8s
kube-system   cilium-sf265                       1/1     Running   0          2m49s
kube-system   cilium-tbmjz                       1/1     Running   0          2m59s
kube-system   coredns-66579b5b88-zsbjv           1/1     Running   0          3m51s
kube-system   metrics-server-57db9dfb7b-zh7g8    1/1     Running   0          3m51s

PR #40, opened 2024-03-04
[COS] Add `COS` substrate fixture

mateoflorido:
## Overview
Add a `COS` substrate fixture to ease COS integration.

### Rationale
We usually test COS integration using the same K8s as the substrate for the COS stack. However, we would want to mimic the real-world environment. This pull request replicates a COS instance related to the K8s cluster using Cross-Model Integration.

### Module Changes
- Added a Grafana module to verify the Grafana integration.
- Added a Prometheus module which verifies the Prometheus integration, querying the cluster for the known metrics jobs.
- Added a COS substrate module implementation using LXD. It creates a separate network and instance, completely independent from the cluster itself (the cluster is using MicroK8s as the K8s substrate).
- Added tests for Grafana and Prometheus.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -8,14 +8,15 @@ jobs:
   extra-args:
     runs-on: ubuntu-latest
     outputs:
-      args: ${{ steps.set-flags.args }}
+      args: ${{ steps.flags.outputs.args }}
     steps:
       - name: Determine extra args
-        id: set-flags
+        id: flags
+        env:
+          TITLE: ${{ github.event.pull_request.title }}
         run: |
           EXTRA_ARGS="--crash-dump=on-failure"
-          PR_TITLE="${{ github.event.pull_request.title }}"
-          if [[ "$PR_TITLE" == *"[COS]"* ]]; then
+          if [[ "$TITLE" == *"[COS]"* ]]; then
             EXTRA_ARGS="$EXTRA_ARGS --cos"
           fi
           echo "args=$EXTRA_ARGS" >> "$GITHUB_OUTPUT"
@@ -38,7 +39,7 @@ jobs:
     with:
       provider: lxd
       juju-channel: 3.3/stable
-      extra-arguments: ${{ needs.extra-args.outputs.args }}
+      extra-arguments: ${{needs.extra-args.outputs.args}}
       load-test-enabled: false
       zap-enabled: false
       trivy-fs-enabled: true
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -1,10 +1,27 @@
+
 name: Integration tests
 
 on:
   pull_request:
 
 jobs:
+  extra-args:
+    runs-on: ubuntu-latest
+    outputs:
+      args: ${{ steps.set-flags.args }}
+    steps:
+      - name: Determine extra args
+        id: set-flags
+        run: |
+          EXTRA_ARGS="--crash-dump=on-failure"
+          PR_TITLE="${{ github.event.pull_request.title }}"
+          if [[ "$PR_TITLE" == *"[COS]"* ]]; then
+            EXTRA_ARGS="$EXTRA_ARGS --cos"
+          fi
+          echo "args=$EXTRA_ARGS" >> "$GITHUB_OUTPUT"
+
   build-all-charms:
+    needs: [extra-args]
     strategy:
       matrix:
         path:
@@ -13,14 +30,15 @@ jobs:
     uses: ./.github/workflows/build-charm.yaml
     with:
       working-directory:  ${{ matrix.path }}
+
   integration-tests:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
-    needs: [build-all-charms]
+    needs: [build-all-charms, extra-args]
     secrets: inherit
     with:
       provider: lxd
       juju-channel: 3.3/stable
-      extra-arguments: --crash-dump=on-failure
+      extra-arguments: ${{ needs.extra-args.outputs.args }}
       load-test-enabled: false
       zap-enabled: false
       trivy-fs-enabled: true
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -4,16 +4,26 @@
 """Fixtures for charm tests."""
 import asyncio
 import contextlib
+import json
 import logging
+import shlex
+import tempfile
 from dataclasses import dataclass, field
 from itertools import chain
 from pathlib import Path
-from typing import List, Mapping, Optional
+from typing import List, Mapping, Optional, Tuple
 
 import pytest
 import pytest_asyncio
 import yaml
-from pytest_operator.plugin import OpsTest
+from juju.model import Model
+from juju.tag import untag
+from kubernetes import config as k8s_config
+from kubernetes.client import Configuration
+from pytest_operator.plugin import OpsTest, ops_test
+
+from .cos_substrate import LXDSubstrate
+from .helpers import get_address
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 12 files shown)


tox.ini
```diff
@@ -84,9 +84,10 @@ commands =
 description = Run integration tests
 deps =
     juju
+    pylxd
     pytest
     pytest-asyncio
-    pytest-operator
+    pytest-operator @ git+https://github.com/charmed-kubernetes/pytest-operator@akd/add-k8s-alpha
```
addyess, line 90: how are we feeling about this now?

tests/integration/test_k8s.py
```diff
@@ -83,6 +86,48 @@ async def test_nodes_ready(kubernetes_cluster: model.Model):
...
+    traefik_address: str,
+    grafana_password: str,
+    expected_dashboard_titles: set,
+    cos_model: model.Model,
+):
+    """Test integration with Grafana."""
+    grafana = Grafana(model_name=cos_model.name, host=traefik_address, password=grafana_password)
+    while not await grafana.is_ready():
+        log.info("Waiting for Grafana to be ready.")
+        await asyncio.sleep(5)
```
addyess, line 100: So you could use `wait_for` here i think -- but it wouldn't have the sleep and log.  maybe this is okay -- but there's no timeout

```suggestion
    await asyncio.wait_for(grafana.is_ready(), timeout=5*60)
```

tests/integration/prometheus.py
```diff
@@ -0,0 +1,87 @@
...
+# See LICENSE file for licensing details.
+
+import json
+import urllib.parse
+import urllib.request
+from typing import Optional
+
+
+
+class Prometheus:
```
addyess, line 11: this is so beautiful

tests/integration/data/cos-offers-overlay.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+applications:
```
addyess, line 4: 😍 i'm in love with this

tests/integration/conftest.py
```diff
@@ -191,3 +201,161 @@ async def kubernetes_cluster(request: pytest.FixtureRequest, ops_test: OpsTest):
...
+        await model.remove_application("grafana-agent")
+
+
+@pytest_asyncio.fixture(scope="module")
+async def cos_model(ops_test: OpsTest, kubernetes_cluster: Model, grafana_agent):
+    """Create a COS substrate and a K8s model."""
+    log.info("Creating COS substrate...")
+    container_name = "cos-substrate"
+    network_name = "cos-network"
+    manager = LXDSubstrate(container_name, network_name)
```
addyess, line 226: hold up... are you saying that to test cos -- you must test with microk8s on lxd?  there's no other option?

tests/integration/conftest.py
```diff
@@ -191,3 +201,161 @@ async def kubernetes_cluster(request: pytest.FixtureRequest, ops_test: OpsTest):
...
+
+    config = manager.create_substrate()
+    kubeconfig_path = ops_test.tmp_path / "kubeconfig"
+    kubeconfig_path.write_text(config)
+    config = type.__call__(Configuration)
+    k8s_config.load_config(client_configuration=config, config_file=str(kubeconfig_path))
+
+    k8s_cloud = await ops_test.add_k8s(kubeconfig=config, skip_storage=False)
```
addyess, line 234: nice, this looks pretty clean

tests/integration/conftest.py
```diff
@@ -191,3 +201,161 @@ async def kubernetes_cluster(request: pytest.FixtureRequest, ops_test: OpsTest):
...
+    k8s_model = await ops_test.track_model(
+        "cos", cloud_name=k8s_cloud, keep=ops_test.ModelKeep.NEVER
+    )
+    with ops_test.model_context("cos"):
+        yield k8s_model
```
addyess, line 239: This should suffice.  No need to switch into the model_context.  It's kind of a convenience method to make things like `ops_test.model` and `ops_test.full_model_name` work for a selected model.  Think of `model_context` like juju switch.  when you `juju status` without a `-m`, you get the currently selected model.  But if you have the model_name, you can always `juju status -m $model`.
```suggestion
   yield k8s_model
```

tests/integration/conftest.py
```diff
@@ -191,3 +201,161 @@ async def kubernetes_cluster(request: pytest.FixtureRequest, ops_test: OpsTest):
...
+    """Deploy Grafana Agent."""
+    with ops_test.model_context("main") as model:
+        await model.deploy("grafana-agent", channel="stable")
+        await model.integrate("grafana-agent:cos-agent", "k8s:cos-agent")
+        await model.integrate("grafana-agent:cos-agent", "k8s-worker:cos-agent")
+        await model.integrate("k8s:cos-worker-tokens", "k8s-worker:cos-tokens")
+
+        yield
+
+        await model.remove_application("grafana-agent")
```
addyess, line 217: even easier

```suggestion
    await kubernetes_cluster.deploy("grafana-agent", channel="stable")
    await kubernetes_cluster.integrate("grafana-agent:cos-agent", "k8s:cos-agent")
    await kubernetes_cluster.integrate("grafana-agent:cos-agent", "k8s-worker:cos-agent")
    await kubernetes_cluster.integrate("k8s:cos-worker-tokens", "k8s-worker:cos-tokens")

    yield

    await kubernetes_cluster.remove_application("grafana-agent")
```

tests/integration/cos_substrate.py
```diff
@@ -0,0 +1,261 @@
...
+            container = self.client.instances.create(config, wait=True)
+            log.info("Starting Container")
+            container.start(wait=True)
+            time.sleep(60)
+            return container
+
+        except Exception as e:
+            log.error(f"Failed to create or start container: {e}")
+            return None
+
```
addyess, line 104: alright -- but is there any chance we can just use our current juju to create another model for microk8s?  Even if that model is holding a bare machine (not necessarily the microk8s charm).


models in this controller would be:
1) controller model
2) test_k8s-1234 <-- model under test
3) k8s-holding-cos-1234 <-- model holding the kubernetes machine used for cos
    4) cos-1234 <--model atop the `add_k8s` from `#3`

addyess (changes requested):
I know it's late and probably a pain, but maybe you can drop `pylxd` if you just use juju to create another model to hold your cos-on-k8s machine for hosting the kubernetes

I hope it's really clear to all who read this PR that we've got 2 independent kubernetes models here. One for hosting apps that watch the other one.

tox.ini
mateoflorido: It's awesome, it simplifies the add-k8s logic a lot!

tests/integration/conftest.py
mateoflorido: Yes, I attempted to run the COS stack on CK. However, the charms are not coming up due to some issues with applying container limits. I strongly believe this is the way to test COS because we are using an external machine where our COS stack is running (probably on another subnet as well).

addyess: what about COS on uK8s -- just running from a machine deployed via juju rather than pylxd? 

some developers may wish to use their cloud resources to test rather than the the local machine where the tests run from. 

#team-inspiron

tox.ini
addyess: So we can merge with this in here.  I'll make a card to release this

.github/workflows/integration_test.yaml
```diff
@@ -1,10 +1,28 @@
...
+      - name: Determine extra args
+        id: flags
+        env:
+          TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          EXTRA_ARGS="--crash-dump=on-failure"
+          if [[ "$TITLE" == *"[COS]"* ]]; then
+            EXTRA_ARGS="$EXTRA_ARGS --cos"
+          fi
+          echo "args=$EXTRA_ARGS" >> "$GITHUB_OUTPUT"
```
addyess, line 22: So, this is clever.  if COS is in the title we run the cos tests.  nice

addyess (approved):
LGTM Professor Florido

.github/workflows/integration_test.yaml
mateoflorido: That's correct, thanks!!

PR #41, opened 2024-03-06
Enable `k8s-operator` to run in proxied environments

mateoflorido:
## Overview

Add a mechanism to create a suitable configuration in `/etc/environment` for Kubernetes to run in a proxied environment.

### Rationale

Currently, the charms fail to start due to issues with the proxy environment variables. This pull request introduces a mechanism to offload the JUJU_CHARM environment variables to the `/etc/environment` file. This would create a suitable environment in which the snap can infer the values and start correctly.

### DIFF
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -16,6 +16,7 @@ certificate storage.
 """
 
 import logging
+import os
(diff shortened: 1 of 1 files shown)


addyess (approved):
Super Duper

PR #43, opened 2024-03-08
Configuration for service-cidr and control-plane-taints

addyess:
Introduces config `service-cidr` and `register-with-taints`.  I considered also introducing `cidr` -- but may wait and leave it default for now in case we later introduce CNI relation

Pairs with snap changes
* https://github.com/canonical/k8s-snap/pull/218


eaudetcobello:
LGTM.

eaudetcobello (approved):
LGTM. Docstrings are clear as well.

PR #47, opened 2024-03-18
Implement External Datastore Integration (ETCD)

mateoflorido:
## Overview
Implement charm integration with external datastores, i.e., ETCD

### DIFF
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -28,9 +28,12 @@ from typing import Dict, Optional
 
 import charms.contextual_status as status
 import ops
+import yaml
 from charms.contextual_status import WaitingStatus, on_error
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from charms.k8s.v0.k8sd_api_manager import (
+    BootstrapConfig,
+    CreateClusterRequest,
     DNSConfig,
     InvalidResponseError,
     K8sdAPIManager,
@@ -40,6 +43,7 @@ from charms.k8s.v0.k8sd_api_manager import (
     UpdateClusterConfigRequest,
     UserFacingClusterConfig,
 )
+from charms.kubernetes_libs.v0.etcd import EtcdReactiveRequires
 from charms.node_base import LabelMaker
 from charms.operator_libs_linux.v2.snap import SnapError, SnapState
 from charms.operator_libs_linux.v2.snap import ensure as snap_ensure
@@ -101,6 +105,8 @@ class K8sCharm(ops.CharmBase):
                 self.on.upgrade_charm,
             ],
         )
+
+        self.etcd = EtcdReactiveRequires(self)
 
         self.framework.observe(self.on.update_status, self._on_update_status)
         if self.is_control_plane:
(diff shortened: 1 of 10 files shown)


tests/integration/test-bundle-etcd.yaml
```diff
@@ -0,0 +1,28 @@
+# Copyright 2024 Canonical Ltd.
```
addyess, line 1: can this go in `tests/integration/data/`

charms/worker/k8s/src/charm.py
```diff
@@ -232,6 +250,27 @@ def _configure_cos_integration(self):
...
+        """Configure etcd as the datastore for the Kubernetes cluster.
+
+        Args:
+            config (BootstrapConfig): The bootstrap configuration object for
+                the Kubernetes cluster that is being configured. This object
+                will be modified in-place to include etcd's configuration details.
+        """
+        etcd_relation = self.model.get_relation("etcd")
+        if not etcd_relation:
+            return
```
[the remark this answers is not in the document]
mateoflorido: During the test runs, I have noticed that the leader waits for the etcd cluster to become available. I wonder if this is a change in the current ops. It seems that the relation object is present in the environment, but with no units in the relation. I think if we have a relation to etcd, but no units yet, the object will be present but with an unready relation.

tests/integration/test-bundle-etcd.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+name: integration-test-etcd
+description: |-
+  Used to deploy or refresh within an integration test model
+series: jammy
+applications:
+  easyrsa:
+    charm: easyrsa
+    channel: stable
+    num_units: 1
```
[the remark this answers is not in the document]
mateoflorido: Yes, just for etcd

charms/worker/k8s/src/charm.py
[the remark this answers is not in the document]
addyess: you're right that it just needs "etcd-relation-created" for the relation to exist.  I'm just afraid that hook doesn't come before the cluster is bootstrapped.  Something has me concerned the cluster would bootstrap before this hook and the relation wouldn't have been created yet.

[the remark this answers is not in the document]
neoaggelos: > you're right that it just needs etcd-relation-created for the relation to exist. I'm just afraid that hook doesn't come before the cluster is bootstrapped.

Right, juju will first do a cycle of `install`, `config-changed` hooks, then fire the `*-relation-created` events. In this scenario, during the initial `install` hook we cannot know if there's an etcd relation waiting to be established, therefore our easiest way out (at the moment) is a config option.

To be fair, I have not tested this with more recent juju versions, so this might have been resolved. However:
- we should not rely on this behaviour if it's not explicitly part of the juju "contract". Even if the _timings_ are such that _99.999%_ percent it goes like we want it to, it would hurt as a lot to design such a race condition that leads to DoA clusters (or, even worse, clusters with an unpredictable state)
- even if it becomes standard juju behaviour in the future, i would still be wary to use it because it would not work with older juju versions, and that's a very dangerous path to tread on.

I think it's preferable for us to be able to control that the charm behaves _exactly_ the way that we want it to.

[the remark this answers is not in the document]
addyess: Maybe charm config like this?  the charm can be deployed via bundle or cli -- and in both cases the config is available on the install hook

```suggestion
        etcd_external = self.config.get("etcd-external")
        if not etcd_external:
            log.info("Charm config allows for bootstrap without an external etcd")
            return
        etcd_relation = self.model.get_relation("etcd")
```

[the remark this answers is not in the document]
mateoflorido: Thank you for your comments, @addyess and @neoaggelos. I have introduced a new configuration option that should resolve this issue. I have opted for a configuration option for the name of the `datastore`, with `dqlite` being the default one.

charms/worker/k8s/src/charm.py
```diff
@@ -248,6 +253,39 @@ def _configure_cos_integration(self):
...
+                will be modified in-place to include etcd's configuration details.
+        """
+        datastore = self.config.get("datastore")
+
+        if datastore not in SUPPORTED_DATASTORES:
+            log.error(
+                "Invalid datastore: %s. Supported values: %s",
+                datastore,
+                ", ".join(SUPPORTED_DATASTORES),
+            )
```
addyess, line 271: You could have the charm block here waiting for a config change

```suggestion
        if datastore not in SUPPORTED_DATASTORES:
            status.add(ops.BlockedStatus(f"Invalid datastore: {datastore}"))
       assert datastore in SUPPORTED_DATASTORES
```

addyess (changes requested):
Small improvement to help the charm block on invalid datastore configuration

charms/worker/k8s/src/charm.py
```diff
@@ -269,6 +269,7 @@ def _configure_datastore(self, config: BootstrapConfig):
                 datastore,
                 ", ".join(SUPPORTED_DATASTORES),
             )
+            status.add(ops.BlockedStatus(f"Invalid datastore: {datastore}"))
```
addyess, line 272: if you don't raise an error here... the reconciler will continue -- deploying with the "dqlite" datastore

tests/integration/test_etcd.py
```diff
@@ -0,0 +1,71 @@
...
+        log.info("Node %s is %s..", node, "ready" if ready else "not ready")
+        assert ready, f"Node not yet ready: {node}."
+
+
+@pytest.mark.abort_on_fail
+async def test_nodes_ready(kubernetes_cluster: model.Model):
+    """Deploy the charm and wait for active/idle status."""
+    k8s = kubernetes_cluster.applications["k8s"]
+    expected_nodes = len(k8s.units)
+    await ready_nodes(k8s.units[0], expected_nodes)
```
addyess, line 71: shouldn't you be expecting 2 nodes,  1 k8s and 1 k8s-worker

tests/integration/data/test-bundle-etcd.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: integration-test-etcd
+description: |-
+  Used to deploy or refresh within an integration test model
+series: jammy
```
addyess, line 7: can we test with focal?
```suggestion
series: focal
```

prev-lxd.yaml
```diff
@@ -0,0 +1,30 @@
...
+    source: /dev/kmsg
+    type: unix-char
+  aadisable3:
+    path: /sys/fs/bpf
+    source: /sys/fs/bpf
+    type: disk
+  aadisable4:
+    path: /proc/sys/net/netfilter/nf_conntrack_max
+    source: /proc/sys/net/netfilter/nf_conntrack_max
+    type: disk
```
addyess, line 30: @mateoflorido i don't understand why this is here?  Is this necessary

mateoflorido: Sorry! A `git add .` mistake 😛

addyess (approved):
LGTM

PR #49, opened 2024-03-22
Remove lxd-profile.yaml from charms, but test with a well known profile

addyess:
Removes the lxd-profile.yaml from the k8s and k8s-worker charms, but uses a single profile during integration tests if the substrate is lxd


tests/integration/conftest.py
```diff
@@ -158,6 +158,17 @@ def switch(self, name: str, path: Path):
         app["channel"] = None
 
 
+async def cloud_profile(ops_test: OpsTest):
+    """Apply lxd-profile to the model if the juju cloud is lxd."""
+    controller = await ops_test.model.get_controller()
+    cloud = await controller.cloud()
+    if cloud.cloud.type_ == "lxd":
+        lxd = LXDSubstrate(None, None)
```
mateoflorido, line 166: Nice!

mateoflorido (approved):
LGTM!

charms/worker/lxd-profile.yaml
mateoflorido, line 1: We need to remove the profile in `charms/worker/k8s/lxd-profile.yaml` as well, right?

addyess: hooray!  it's been MOVED to `/tests/integration/data/k8s.profile`

PR #50, opened 2024-03-22
Unify all the tox jobs into the top-level

addyess:
Unifies to a single pyproject.toml and applies linter rules to all `.py` files

Depends on #51 to be first merged


addyess:
@louiseschmidtgen I got around to removing all those extraneous `pyproject.toml` files

.github/workflows/tests-k8s.yaml
```diff
@@ -1,4 +1,5 @@
-name: Tests K8s
+name: Charm Analysis
```
louiseschmidtgen, line 1: Should we change the name of the file to `charm-analysis.yaml` and keep things consistent?

louiseschmidtgen (approved):
Thanks for cleaning up all these files,looks like we cover the COS code as well now! Nice!
Overall, LGTM I only have a rename suggestion.

.github/workflows/tests-k8s.yaml
addyess: as you wish

PR #52, opened 2024-03-25
Expose ports and apps when running in public clouds

addyess:
Expose ports to the kubernetes api endpoint service on control-plane units, and expose those apps on public clouds during testing

Without this, security groups will prevent using kubernetes clients from accessing the API endpoints outside the cloud


bschimke95 (approved):
LGTM

PR #53, opened 2024-03-25
CI timeout

louiseschmidtgen:
## Description
Set CI timeout 120 min.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -33,6 +33,7 @@ jobs:
       working-directory:  ${{ matrix.path }}
 
   integration-tests:
+    timeout-minutes: 120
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -6,6 +6,7 @@ on:
 
 jobs:
   extra-args:
+    timeout-minutes: 120
     runs-on: ubuntu-latest
```
addyess, line 10: this times the `extra-args` job.  You wanna time the `integration-tests` job on line 36

louiseschmidtgen: True. Maybe my Vscode is lying let's see what CI says...

louiseschmidtgen:
Needs to be set in operator-workflows. New PR: https://github.com/canonical/operator-workflows/pull/267

PR #56, opened 2024-03-28
include license check on k8s charm library but ignore downloaded library

addyess:
This charm maintains the `k8s` charm library under `charm/worker/k8s/lib/charms/k8s` so we should license check those files.  But we want to ignore any other charm license headers bc we can't do anything about them


bschimke95 (approved):
LGTM

PR #59, opened 2024-04-01
Use self-hosted runners

addyess:
Attempt to use a "large" self-hosted runner for integration tests


bschimke95 (commented):
Seems like the cdk bot does not have sufficient permissions for this.

addyess:
Agreed, it was just a test to see what was blocking us from self-hosted runners.  thanks @bschimke95

mateoflorido (approved):
LGTM!

PR #63, opened 2024-04-10
Adjust to updated operator-workflows

addyess:
https://github.com/canonical/operator-workflows/pull/268 Adjusted the operator-workflows in a non-backwards compatible way.

This fixes our workflows to use the new inputs


mateoflorido (approved):
LGTM!

PR #65, opened 2024-04-12
Ensure all bootstrap arguments are nil unless otherwise specified

addyess:
Summary:
Pydantic models do not need to populate fields that can be sent to the server as nil.  It's best to omit them to prevent unnecessary server-side checking. 

Changes:
* Change the Models to default all fields to `None`
* Adjust the linter to check the `...k8s/lib/charms/k8s` path
* Adjust the tests to set constraints on the machines for CI (in the event they are not lxd -- these constraints are applied)
* Add integration test to confirm we're using external datastore
* Wrap tokens in a `SecretStr` to prevent the secret from leaking in logs


mateoflorido (approved):
LGTM

PR #66, opened 2024-04-14
Handle tagging with different prefix during charm publish

addyess:
When publishing the charms to latest/edge, tag the repo with multiple tags, one for each charm revision.


addyess:
@bschimke95 I think this fixes what you spotted last week on merges to publishes to charmhub.

mateoflorido (approved):
LGTM!

PR #68, opened 2024-04-16
Support the external-cloud-provider on bootstrap of k8s

addyess:
Satisfies the `external-cloud-provider` relation - the k8s charms can bootstrap using the `external` cloud-provider

While this isn't enough for the other cloud-provider charms to get credentials from the cluster -- this is enough to complete half the relation and indicate that the cloud-provider is externally available. 

--- Note ---
`aws-cloud-provider` requires the nodes to use fqdn rather than the bare hostname in the node names.  It's best in this case to just always use the fqdn when running on aws

### DIFF
--- a/charms/worker/k8s/charmcraft.yaml
+++ b/charms/worker/k8s/charmcraft.yaml
@@ -98,3 +98,5 @@ provides:
 requires:
   etcd:
     interface: etcd
+  external-cloud-provider:
+    interface: external_cloud_provider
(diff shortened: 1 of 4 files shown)


louiseschmidtgen (commented):
Hi @addyess,

We will have to update the cloud-provider in userfacingclusterconfig and bootstrap config since I merged https://github.com/canonical/k8s-snap/pull/344.

Apologies, only realized I broke things here after I hit merge.

kwmonroe (approved):
Nice job adam, xcp ftw!

PR #71, opened 2024-04-18
Integrate runtime updates for etcd

bschimke95:
## Overview

With https://github.com/canonical/k8s-snap/pull/334 the snap allows to configure the external datastore on runtime. This allows for changes of the etcd configuration after the cluster is bootstrapped. 
The snap only performs steps (restarting services etc.) if a configuration changed. Hence, we can just send the current datastore configuration of the charm to the endpoint on every event. If there is no change, nothing happens.

## Changes

* Add `UserFacingDatastoreConfig` type as defined in the `k8s-snap` API
* Update `Datastore` status type
* Add `datastore` to the `UpdateClusterConfigRequest`
* Add new reconcile step `ensure_cluster_config`

### DIFF
--- a/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
+++ b/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
@@ -304,7 +304,7 @@ class UserFacingDatastoreConfig(BaseModel, allow_population_by_field_name=True):
         client_key: client key of the external datastore cluster in PEM format.
     """
 
-    type: Optional[str]] = Field(None)
+    type: Optional[str] = Field(None)
     servers: Optional[List[str]] = Field(None)
     ca_crt: Optional[str] = Field(None, alias="ca-crt")
     client_crt: Optional[str] = Field(None, alias="client-crt")
--- a/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
+++ b/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
@@ -290,6 +290,24 @@ class UserFacingClusterConfig(BaseModel):
     metrics_server: MetricsServerConfig = Field(None, alias="metrics-server")
 
 
+class UserFacingDatastoreConfig(BaseModel):
+    """Aggregated configuration model for the user-facing datastore aspects of a cluster.
+
+    Attributes:
+        type: Type of the datastore. For runtime updates, this needs to be "external".
+        servers: Server addresses of the external datastore.
+        ca_crt: CA certificate of the external datastore cluster in PEM format.
+        client_crt: client certificate of the external datastore cluster in PEM format.
+        client_key: client key of the external datastore cluster in PEM format.
+    """
+
+    type: str = Field(None)
+    servers: List[str] = Field(None)
+    ca_crt: str = Field(None, alias="ca-crt")
+    client_crt: str = Field(None, alias="client-crt")
+    client_key: str = Field(None, alias="client-key")
+
+
 class BootstrapConfig(BaseModel):
     """Configuration model for bootstrapping a Canonical K8s cluster.
 
@@ -341,9 +359,11 @@ class UpdateClusterConfigRequest(BaseModel):
 
     Attributes:
         config (Optional[UserFacingClusterConfig]): The cluster configuration.
+        datastore (Optional[UserFacingDatastoreConfig]): The clusters datastore configuration.
     """
 
-    config: UserFacingClusterConfig
+    config: UserFacingClusterConfig = Field(None)
+    datastore: UserFacingDatastoreConfig = Field(None)
 
 
 class DatastoreStatus(BaseModel):
@@ -351,11 +371,11 @@ class DatastoreStatus(BaseModel):
 
     Attributes:
         datastore_type (str): external or k8s-dqlite datastore
-        external_url: (str): list of external_urls
+        servers: (List(str)): list of server addresses of the external datastore cluster.
     """
 
     datastore_type: str = Field(None, alias="type")
-    external_url: str = Field(None, alias="external-url")
+    servers: List[str] = Field(None, alias="servers")
 
 
 class ClusterStatus(BaseModel):
--- a/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
+++ b/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
@@ -290,7 +290,7 @@ class UserFacingClusterConfig(BaseModel):
     metrics_server: MetricsServerConfig = Field(None, alias="metrics-server")
 
 
-class UserFacingDatastoreConfig(BaseModel, allow_population_by_field_name=True):
+class UserFacingDatastoreConfig(BaseModel, allow_population_by_field_name=True):  # type: ignore[call-arg]
     """Aggregated configuration model for the user-facing datastore aspects of a cluster.
 
     Attributes:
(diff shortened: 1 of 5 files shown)


charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
```diff
@@ -262,13 +263,29 @@ def test_update_cluster_config(self, mock_send_request):
 
         dns_config = DNSConfig(enabled=True)
         user_config = UserFacingClusterConfig(dns=dns_config)
-        request = UpdateClusterConfigRequest(config=user_config)
+        datastore = UserFacingDatastoreConfig(
+            type="external",
+            servers=["localhost:123"],
+            ca_crt="ca-crt",
```
bschimke95, line 269: for some reason, this fails. I set the alias for those fields but it will only accept `ca-crt` and ignores `ca_crt`?

tests/integration/test_etcd.py
```diff
@@ -40,4 +40,6 @@ async def test_etcd_datastore(kubernetes_cluster: model.Model):
     status = json.loads(result.results["stdout"])
     assert status["ready"], "Cluster isn't ready"
     assert status["datastore"]["type"] == "external", "Not bootstrapped against etcd"
-    assert status["datastore"]["external-url"] == f"https://{etcd.public_address}:{etcd_port}"
+    assert status["datastore"]["servers"] == [
```
bschimke95, line 43: how can we test that the datastore config update works as expected? @mateoflorido is there an easy way to add an etcd member while the test is running?

mateoflorido: You can add a new test in the `test_etcd.py` module; something like `def test_update_etcd_cluster()`. For example, you could do something like [this](https://github.com/canonical/k8s-operator/blob/main/tests/integration/test_k8s.py#L98). Get the application from the model, and add an unit. Then you should verify that your etcd server's string has been updated.

charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
mateoflorido: What's the problem here? I think I don't follow... Did you mean that the model is serialized using `ca_crt` as the JSON key instead of `ca-crt`.

mateoflorido: I see it now: 
```
AssertionError: expected call not found.
Expected: _send_request('/1.0/k8sd/cluster/config', 'PUT', <class 'lib.charms.k8s.v0.k8sd_api_manager.EmptyResponse'>, {'config': {'dns': {'enabled': True}}, 'datastore': {'type': 'external', 'servers': ['localhost:123'], 'ca-crt': 'ca-crt', 'client-crt': 'client-crt', 'client-key': 'client-key'}})
Actual: _send_request('/1.0/k8sd/cluster/config', 'PUT', <class 'lib.charms.k8s.v0.k8sd_api_manager.EmptyResponse'>, {'config': {'dns': {'enabled': True}}, 'datastore': {'type': 'external', 'servers': ['localhost:123']}})
```

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -290,6 +290,24 @@ class UserFacingClusterConfig(BaseModel):
     metrics_server: MetricsServerConfig = Field(None, alias="metrics-server")
 
 
+class UserFacingDatastoreConfig(BaseModel):
```
addyess, line 293: Thanks to @mateoflorido for this suggestion
```suggestion
class UserFacingDatastoreConfig(BaseModel, allow_population_by_field_name=True):
```

addyess (changes requested):
Hopefully some suggestions for the unit tests to pass

charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
bschimke95: yes, `allow_population_by_field` was the solution, thanks!

tests/integration/test_etcd.py
bschimke95: Thanks, added an integration test with this.

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
bschimke95: Thanks, that fixed it!

tests/integration/test_etcd.py
```diff
@@ -40,4 +40,26 @@ async def test_etcd_datastore(kubernetes_cluster: model.Model):
...
+
+
+@pytest.mark.abort_on_fail
+async def test_update_etcd_cluster(kubernetes_cluster: model.Model):
+    """Test that adding etcd clusters are propagated to the k8s cluster."""
+    k8s: unit.Unit = kubernetes_cluster.applications["k8s"].units[0]
+    etcd = kubernetes_cluster.applications["etcd"]
+
+    await etcd.add_unit()
+    await etcd.add_unit()
```
addyess, line 53:
```suggestion
    await etcd.add_unit(count=2)
    await model.wait_for_idle(status="active")
```

addyess, line 53: Make sure there are at least 3 etcd units.  If you use this technique, you can re-run the same test over and over without it adding lots of etcd units.

```suggestion
    count = 3 - len(etcd.units)
    if count > 0:
        await etcd.add_unit(count=count)
    await model.wait_for_idle(status="active")
```

addyess (approved):
Looking really good -- one suggested change

charms/worker/k8s/src/charm.py
```diff
@@ -361,6 +362,36 @@ def _enable_functionalities(self):
...
+            etcd_config = self.etcd.get_client_credentials()
+            update_request.datastore = UserFacingDatastoreConfig(
+                type="external",
+                servers=self.etcd.get_connection_string().split(","),
+                ca_crt=etcd_config.get("client_ca", ""),
+                client_crt=etcd_config.get("client_cert", ""),
+                client_key=etcd_config.get("client_key", ""),
+            )
+
+        self.api_manager.update_cluster_config(update_request)
```
addyess, line 393: I continue to encounter this error when the third etcd unit tries to post it's update:

```
charms.k8s.v0.k8sd_api_manager.InvalidResponseError: Error status 500
	method=PUT
	endpoint=/1.0/k8sd/cluster/config
	reason=Internal Server Error
	body={"type":"error","status":"","status_code":0,"operation":"","error_code":500,"error":"failed to reconcile network: failed to refresh network: failed to refresh network component: failed to upgrade component 'network': another operation (install/upgrade/rollback) is in progress","metadata":null}
```

addyess: Here are the k8sd and charm logs around the same moment of this failure
https://pastebin.canonical.com/p/jD4TNbk4j4/

addyess: api server logs around the same moment
https://pastebin.canonical.com/p/QzMpFCmQCV/

is it possible that adding extra etcd units in rapid succcession caused the API server to restart when the first datastore came online, and when we published the second datastore URL -- the config wouldn't take?

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -346,9 +364,11 @@ class UpdateClusterConfigRequest(BaseModel):
 
     Attributes:
         config (Optional[UserFacingClusterConfig]): The cluster configuration.
+        datastore (Optional[UserFacingDatastoreConfig]): The clusters datastore configuration.
     """
 
-    config: UserFacingClusterConfig
+    config: UserFacingClusterConfig = Field(None)
+    datastore: UserFacingDatastoreConfig = Field(None)
```
addyess, line 371:
```suggestion
    config: Optional[UserFacingClusterConfig] = Field(None)
    datastore: Optional[UserFacingDatastoreConfig] = Field(None)
```

charms/worker/k8s/src/charm.py
bschimke95: Hey Adam,
Thanks for the review. I think the issue you were facing was before https://github.com/canonical/k8s-snap/pull/356 was merged. 

I cannot reproduce this issue and the CI also seems to be happy.

PR #73, opened 2024-04-19
Join nodes and Bootstrap nodes with extra-sans config

addyess:
### Overview

On Bootstrap and on Join Node -- we should pass extra configuration about the node on the cluster --namely the extra-sans argument. 

### Rationale

With extra-sans, the generated kube-config will allow for the certificate to share the SubjectAltNames enabling secure TLS

### Library Changes

Updated the types of many of the structures used in the api

### DIFF
--- a/tests/integration/test_etcd.py
+++ b/tests/integration/test_etcd.py
@@ -40,4 +40,4 @@ async def test_etcd_datastore(kubernetes_cluster: model.Model):
     status = json.loads(result.results["stdout"])
     assert status["ready"], "Cluster isn't ready"
     assert status["datastore"]["type"] == "external", "Not bootstrapped against etcd"
-    assert status["datastore"]["external-url"] == f"https://{etcd.public_address}:{etcd_port}"
+    assert status["datastore"]["servers"] == [f"https://{etcd.public_address}:{etcd_port}"]


charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -346,6 +349,68 @@ class UpdateClusterConfigRequest(BaseModel):
...
+            kwds: keyword arguments
+
+        Returns:
+            dict mapping of the object
+        """
+        rendered = super().dict(**kwds)
+        rendered["token"] = self.token.get_secret_value()
+        if self.config:
+            rendered["config"] = yaml.safe_dump(self.config.dict(**kwds))
+        return rendered
```
addyess, line 411: override the built-in pydantic `dict(...)` method to render the `config` argument as yaml, and the `token` as a secret string

charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
```diff
@@ -233,11 +235,36 @@ def test_create_join_token_worker(self, mock_send_request):
...
+        self.api_manager.join_cluster(request)
+        mock_send_request.assert_called_once_with(
+            "/1.0/k8sd/cluster/join",
+            "POST",
+            EmptyResponse,
+            {
+                "name": "test-node",
+                "address": "127.0.0.1:6400",
+                "token": "test-token",
+                "config": "extra-sans:\n- 127.0.0.1\n",
```
addyess, line 255: oo look yaml

pyproject.toml
```diff
@@ -40,6 +40,7 @@ profile = "black"
 ignore_missing_imports = true
 explicit_package_bases = true
 namespace_packages = true
+plugins = "pydantic.mypy"
```
addyess, line 43: NIFTY!  Now pydantic types are type checked

mateoflorido (approved):
Awesome! LGTM 🥇

PR #78, opened 2024-04-23
Cache Charm Builds

mateoflorido:
## Overview
This pull request introduces a caching mechanism for charms. If there are no changes in the source of the charm or its dependencies, the workflow restores the cached charms. This reduces integration test times by around 80%. This is particularly useful when tweaking the testing suite or other files that are not dependent on the charm sources.

### DIFF
--- a/.github/workflows/build-charm.yaml
+++ b/.github/workflows/build-charm.yaml
@@ -34,8 +34,14 @@ jobs:
         id: cache-charm
         uses: actions/cache@v4
         with:
-          path: ${{ inputs.working-directory }}/packed/
+          path: ${{ inputs.working-directory }}packed/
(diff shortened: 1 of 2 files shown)


addyess:
so, what happens if one of the deps in requirements.txt changes?  The hash would still match right?  How do we pick up new changes of python dependencies?

mateoflorido:
> so, what happens if one of the deps in requirements.txt changes? The hash would still match right? How do we pick up new changes of python dependencies?

No, we are hashing the `.py`, `.yaml` and the `requirements.txt` files in the source folder. If we change any deps in the requirements file, the hash won't match and the charms will be rebuilt.

addyess (approved):
Hopefully dependabot will help out here updating these packages. 
I dunno if it knows how to update the `git+https` packages

PR #80, opened 2024-04-25
Improve node removal by validating departed|broken relations

addyess:
## Summary
Attempts to lengthen the amount of time that a leaving node exists to give time to the lead control-plane charm time to remove the unit

## Changes
* revoke tokens only when `k8s-cluster` and `cluster` relations experience a RelationDepartedEvent
* detect departing unit and elect for removal
* `update_status` event now checks if the node itself is ready with kubectl
* `last_gasp` now requires the node to be reported not in a `Ready` at least 3 times before ending the delay

### DIFF
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -631,7 +631,7 @@ class K8sCharm(ops.CharmBase):
         node = node or self.get_node_name()
         cmd = ["nodes", node, '-o=jsonpath={.status.conditions[?(@.type=="Ready")].status}']
         try:
-            return self.kubectl_get(cmd) == "True"
+            return self.kubectl_get(*cmd) == "True"
         except subprocess.CalledProcessError:
             return False
 
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -639,10 +639,10 @@ class K8sCharm(ops.CharmBase):
         """Busy wait on stop event until the unit isn't clustered anymore."""
         busy_wait, reported_down = 30, 0
         status.add(ops.MaintenanceStatus("Ensuring cluster removal"))
-        while busy_wait and reported_down == 3:
+        while busy_wait and reported_down != 3:
             log.info("Waiting for this unit to uncluster")
-            if self._is_node_ready():
-                log.info("Cluster Node is still ready")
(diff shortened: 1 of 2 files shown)


charms/worker/k8s/src/charm.py
```diff
@@ -78,6 +78,22 @@ def _get_public_address() -> str:
...
+    Args:
+        event (ops.EventBase): event to consider.
+
+    Returns:
+        Optional[ops.Unit] Unit leaving the cluster
+    """
+    return (
+        isinstance(event, ops.RelationDepartedEvent)
+        and event.relation.name in ["k8s-cluster", "cluster"]
+        and event.departing_unit
```
mateoflorido, line 93: Nice use of the `and` operator 😄

mateoflorido (approved):
LGTM!

charms/worker/k8s/src/charm.py
addyess:
````python
>>> 1 and 2 and 3 and 4
4
```

charms/worker/k8s/src/charm.py
```diff
@@ -567,6 +595,10 @@ def _update_status(self):
         if version := self._get_snap_version():
             self.unit.set_workload_version(version)
 
+        if not self._is_node_ready():
+            status.add(ops.WaitingStatus("Node not yet Ready"))
+            return
```
addyess, line 600: This is actually a nice little check for update-status to keep the node from showing "Ready" when -- actually it can't host any workloads yet because it's not Kubernetes Ready

charms/worker/k8s/src/charm.py
```diff
@@ -618,6 +670,39 @@ def _on_update_status(self, _event: ops.UpdateStatusEvent):
         except status.ReconcilerError:
             log.exception("Can't update_status")
 
+    def kubectl(self, *args) -> str:
```
bschimke95, line 673: The charm file starts to get quite large. Maybe it makes sense to split it up in the future, e.g. by moving such utils into their own classes.

bschimke95 (approved):
LGTM, great work. 
Always fun to learn some new Python magic. (e.g. that `and` chains return the last element)

charms/worker/k8s/src/charm.py
addyess: it is -- and i got REALLY close to pulling the trigger on that here.

PR #83, opened 2024-04-25
github cron job to update release branch snap_installation.yaml

addyess:
## Summary
Create a github cron job which updates release branches with latest available stable revision

## Details
* Creates PRs into `release-xx` branches to update the static revision configuration
* Expects that there is a stable risk snap in the release channel
* Might get confused if there are more than one match track in snapstore with the same revision (eg. 1.30 vs 1.30-classic)


.github/workflows/update-snap-revision.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.12'
+
+    - name: Update Revision
+      id: update-revision
+      run: |
+        python3 .github/workflows/update-snap-revision.py ${{ matrix.arch }} ${{ env.TRACK }} stable
```
mateoflorido, line 54: Just one question. Where is this TRACK envvar set?

addyess: wow, nice call.  Updated and spotted another bug.  Thanks @mateoflorido

.github/workflows/publish-charms.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
           if [[ "${BRANCH}" == "main" ]]; then
             echo "track=latest" >> "$GITHUB_OUTPUT"
             echo "risk=edge" >> "$GITHUB_OUTPUT"
-          elif [[ "${BRANCH}" =~ "^release-[0-9]+\.[0-9]+$" ]]; then
+          elif if [[ "${BRANCH}" =~ ^release-[0-9]+\.[0-9]+$ ]]; then
```
mateoflorido, line 26: Is this correct? I haven't seen this before

addyess: fixed...

.github/workflows/update-snap-revision.yaml
addyess: track env is now from [here](https://github.com/canonical/k8s-operator/pull/83/files#diff-a2827cd6b7ce4ca1549d38bb2c36326ea0662e9e5f874a469569480ab3c9d119R40-R43)

.github/workflows/promote-charms.yaml
```diff
@@ -37,8 +37,11 @@ jobs:
           BRANCH=${BRANCH#refs/heads/}  # strip off refs/heads/ if it exists
           if [[ "${BRANCH}" == "main" ]]; then
             echo "track=latest" >> "$GITHUB_OUTPUT"
-          elif [[ "${BRANCH}" =~ "^release-[0-9]+\.[0-9]+$" ]]; then
+          elif [[ "${BRANCH}" =~ ^release-[0-9]+\.[0-9]+$ ]]; then
             echo "track=${BRANCH:8}" >> "$GITHUB_OUTPUT"
+          else
+            echo "::error Failed to determine track from branch ${BRANCH}"
+          exit 1
```
mateoflorido, line 44: Perhaps another syntax error?
```suggestion
          else
            echo "::error Failed to determine track from branch ${BRANCH}"
            exit 1
          fi
```

Issue #85, opened 2024-04-26
K8s Leader fails to remove node


addyess:
### Bug Description

In CI, i found a k8s unit whose unit was gone, but had not been removed from the cluster.  

Upon further investigation, it seems that the unit was never placed in the k8s cluster relation databag. 
When the node-name is not in this mapping, the cluster leader doesn't know it should be removed by the token_distributor. 

### To Reproduce

```bash
juju deploy k8s -n 3
# wait for stable
juju remove-unit k8s/2
```

### Environment

GH CI

### Relevant log output

```shell
This is the most relevant log from the k8s leader.  You'll noticed the departed relation hook for k8s/2 -- but there's no mention of "Completed Token allocation of ..." for k8s/2


2024-04-26 11:12:10 DEBUG juju.worker.uniter.operation executor.go:85 running operation run relation-departed (0; unit: k8s/2, departee: k8s/2) hook for k8s/1
2024-04-26 11:12:10 DEBUG juju.worker.uniter.operation executor.go:88 acquiring machine lock for k8s/1
2024-04-26 11:12:10 DEBUG juju.machinelock machinelock.go:166 acquire machine lock "machine-lock" for k8s/1 uniter (run relation-departed (0; unit: k8s/2, departee: k8s/2) hook)
2024-04-26 11:12:10 DEBUG juju.machinelock machinelock.go:176 machine lock "machine-lock" acquired for k8s/1 uniter (run relation-departed (0; unit: k8s/2, departee: k8s/2) hook)
2024-04-26 11:12:10 DEBUG juju.worker.uniter.operation executor.go:135 preparing operation "run relation-departed (0; unit: k8s/2, departee: k8s/2) hook" for k8s/1
2024-04-26 11:12:10 DEBUG juju.worker.uniter.operation executor.go:135 executing operation "run relation-departed (0; unit: k8s/2, departee: k8s/2) hook" for k8s/1
2024-04-26 11:12:10 DEBUG juju.worker.uniter agent.go:22 [AGENT-STATUS] executing: running cluster-relation-departed hook for k8s/2
2024-04-26 11:12:10 DEBUG juju.worker.uniter.runner runner.go:719 starting jujuc server  {unix @/var/lib/juju/agents/unit-k8s-1/agent.socket <nil>}
2024-04-26 11:12:10 DEBUG unit.k8s/1.juju-log server.go:325 cluster:0: ops 2.12.0 up and running.
[...]

PR #86, opened 2024-04-26
Handle situations where nodes join the cluster, but fail to report to the leader

addyess:
Addresses https://github.com/canonical/k8s-operator/issues/85

### Overview
* Adjustments to the node joiner to report "joined" on its cluster relation
* Adjustments to the cluster leader to be prepared for units which fail to report "joined"
* Revocation of juju secret once nodes validate they are using the one-shot join token
* Revocation of juju secret once a node LEAVE a cluster (whether the token was used or not)
* Adjustment to the remove-myself logic when the leader is leaving to prevent other relation departed hooks from tearing down workers

### Rationale

Due to snap timeouts on the API, nodes can join a cluster on one charm hook, but they fail to report back to the leader with a "joined" on the cluster relation.  The world moves on -- but that node is set for disaster if it ever tries to LEAVE the cluster.  The application databag won't get updated, so the revoke logic doesn't know the node name to destroy the node later in life.  

### Juju Events Changes

* Swapping out the `cluster` relation's interface name to `k8s-cluster`
* Swapping out the `cos-token` relation's interface name to `cos-k8s-token`

### Checklist

- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)


charms/worker/k8s/src/charm.py
```diff
@@ -351,11 +351,11 @@ def _revoke_cluster_tokens(self, event: ops.EventBase):
...
         log.info("Garbage collect cluster tokens")
-        to_remove = set()
+        to_remove = None
         if self.is_dying:
-            to_remove |= {self.unit}
-        if unit := _cluster_departing_unit(event):
-            to_remove |= {unit}
+            to_remove = self.unit
+        elif unit := _cluster_departing_unit(event):
+            to_remove = unit
```
addyess, line 358: let's just simplify this.  If we're handling a hook b/c this unit is dying... only revoke the cluster tokens for the node specified.  This code is only run on the control-plane leader.  If a non "leaving style" event is handled `to_remove` its empty.  If it's a leaving-style event, then it's the unit that is leaving.  It could be a worker leaving, it could be a control-plane leaving -- and HARDEST -- it could be the lead control-plane leaving

charms/worker/k8s/src/charm.py
```diff
@@ -515,10 +515,12 @@ def _get_snap_version(self) -> Optional[str]:
...
         """Retrieve the join token from secret databag and join the cluster."""
-        if self.is_control_plane and self.api_manager.is_cluster_bootstrapped():
-            return
-
         if not (relation := self.model.get_relation("cluster")):
+            status.add(ops.BlockedStatus("Missing cluster integration"))
+            assert False, "Missing cluster integration"  # nosec
+
+        if self.is_control_plane and self.api_manager.is_cluster_bootstrapped():
+            relation.data[self.unit]["joined"] = self.get_node_name()
```
addyess, line 523: The switch in the order is here b/c `cluster` as a peer relation will ALWAYS exist on control-plane units -- so that's a no-op and if it's not there on the worker -- let's just GIVE Up completely on the reconciliation loop and wait

Next -- this is important.  If i'm a control-plane unit and i think im in a bootstrapped cluster, lets just say i'm joined.  There's no harm here.  it may write the same relation data again -- but if the previous hook joined -- but we got some kind of 500 error from the snap like a context timeout -- then we REALLY did get joined, but just didn't finish the context method in `self.collector.joined(...)`  If we don't tell the leader control-plane that we are joined -- this  node will have a hard time leaving the cluster in the future

charms/worker/k8s/src/charm.py
```diff
@@ -530,14 +532,14 @@ def _join_cluster(self):
...
-            log.info("Joining %s to %s...", node_name, cluster_addr)
+            log.info("Joining %s(%s) to %s...", self.unit, node_name, cluster_addr)
             request = JoinClusterRequest(name=node_name, address=cluster_addr, token=token)
             if self.is_control_plane:
                 request.config = ControlPlaneNodeJoinConfig()
                 request.config.extra_sans = [_get_public_address()]
 
             self.api_manager.join_cluster(request)
-            log.info("Success")
+            log.info("Joined %s(%s)", self.unit, node_name)
```
addyess, line 542: this log message was terrible.

charms/worker/k8s/src/charm.py
```diff
@@ -587,11 +589,7 @@ def _update_status(self):
         if self.is_dying:
             status.add(ops.WaitingStatus("Preparing to leave cluster"))
             return
-        if self.is_worker:
-            if not self.model.get_relation("cluster"):
-                status.add(ops.BlockedStatus("Missing cluster integration"))
-                assert False, "Missing cluster integration"  # nosec
-        else:
+        if self.is_control_plane:
```
addyess, line 592: Cool -- this worker status is now handled in the `join_cluster` method.  We don't need it here anymore.

charms/worker/k8s/src/token_distributor.py
```diff
@@ -164,24 +164,24 @@ def _create_cos_token(self, name: str, _):
             username=f"system:cos:{name}", groups=["system:cos"]
         )
 
-    def _revoke_cluster_token(self, name: str, _):
+    def _revoke_cluster_token(self, name: str, ignore_errors: bool):
```
addyess, line 167: OK, there may be other reasons we want to ignore errors other than just i'm the leader and i want out of the cluster.  For instance -- we can try to revoke the token of a node that half-joined, or failed to join.  This lets us effectively "try" to remove this node from the cluster.  More below

charms/worker/k8s/src/token_distributor.py
```diff
@@ -164,24 +164,24 @@ def _create_cos_token(self, name: str, _):
...
-        except (K8sdConnectionError, InvalidResponseError):
-            if name == self.node_name:
-                # I'm unclustering myself
+        except (K8sdConnectionError, InvalidResponseError) as e:
+            if ignore_errors:
                 # Let's just ignore some of these expected errors:
                 # "Remote end closed connection without response"
                 # "Failed to check if node is control-plane"
-                log.exception("Unclustering ones self can expect an error")
+                log.warning("Unclustering self leads to an expected error: %s", e)
```
addyess, line 184: Having a stack-trace here wasn't useful debug.  
Oo... just noticed this warning isn't quite right either.  There may be other reasons we are ignoring errors.

charms/worker/k8s/src/token_distributor.py
```diff
@@ -193,6 +193,30 @@ def _revoke_cos_token(self, name: str, _):
...
+
+        Args:
+            relation (ops.Relation): Which relation (cluster or k8s-cluster)
+            unit (ops.Unit): The unit the secret is intended for
+        """
+        secret_id = SECRET_ID.format(unit.name)
+        if juju_secret := relation.data[self.charm.unit].pop(secret_id, None):
+            secret = self.charm.model.get_secret(id=juju_secret)
+            secret.remove_all_revisions()
+
```
addyess, line 219: These are here to clean up the tokens we keep making that juju is holding onto forever.  Once a token is USED (or revoked) juju doesn't need to know about that secret anymore.  Let's be a good citizen and tidy up

charms/worker/k8s/src/token_distributor.py
```diff
@@ -245,12 +269,20 @@ def allocate_tokens(
                     relation.name,
                     unit.name,
                 )
-                app_databag[unit.name] = name
-                if invalidate_on_join:
-                    relation.data[self.charm.unit].pop(secret_id, None)
+                app_databag[unit.name] = f"joined-{name}"
+                if revoke_on_join:
+                    self._revoke_juju_secret(relation, unit)
+
```
addyess, line 275: The remote unit is joined.  That means that token was used.  Let's CLEAN up the secret

charms/worker/k8s/src/token_distributor.py
```diff
@@ -245,12 +269,20 @@ def allocate_tokens(
...
             if relation.data[self.charm.unit].get(secret_id):
-                continue  # unit already assigned a token
+                # unit already assigned a token
+                log.info(
+                    "Waiting for token to be recovered %s unit=%s:%s",
+                    token_type.value,
+                    relation.name,
+                    unit.name,
+                )
+                continue
```
addyess, line 285: I wanted a log here to see if the leader control plane was ever seeing it had handed out a token, that hadn't yet been used

charms/worker/k8s/src/token_distributor.py
```diff
@@ -260,13 +292,14 @@ def allocate_tokens(
             secret = relation.app.add_secret(content)
             secret.grant(relation, unit=unit)
             relation.data[self.charm.unit][secret_id] = secret.id or ""
+            app_databag[unit.name] = f"pending-{name}"
```
addyess, line 295: # BIG CHANGE HERE

rather than JUST putting the nodename in the `app_databag`, let's put a state and the nodename

There are two states:
* `pending` meaning that we created a token, and it hasn't been used yet
* `joined` meaning we KNOW the other end used it

charms/worker/k8s/src/token_distributor.py
```diff
@@ -245,12 +269,20 @@ def allocate_tokens(
                     relation.name,
                     unit.name,
                 )
-                app_databag[unit.name] = name
-                if invalidate_on_join:
-                    relation.data[self.charm.unit].pop(secret_id, None)
+                app_databag[unit.name] = f"joined-{name}"
```
addyess, line 272: go look [here](https://github.com/canonical/k8s-operator/pull/86/files#r1581551116)

charms/worker/k8s/src/token_distributor.py
```diff
@@ -275,24 +308,30 @@ def revoke_tokens(
...
         app_databag = relation.data[self.charm.app]
         joined = {self.charm.model.get_unit(str(u)) for u in app_databag}
-        if to_remove is not None:
-            remaining = joined - to_remove
-        else:
-            remaining = joined & (relation.units | {self.charm.unit})
-        remove = joined - remaining
+
+        # establish the remaining units
+        remove = {to_remove} if to_remove else (joined - all_units)
```
addyess, line 327: if i'm told to remove one unit -- only try to remove that ONE unit -- ignore trying to reap all the others for now

addyess, line 327: I feel like you need to have a math minor to deal with the set theory in this change. 

Problem:  There's 2 sets of truths
1) The app_databag keeps up with a mapping of `units` -> `state + nodename`
2) the relation databag which tells us the current relation status
   * caveat:  on relation-departed hooks -- the unit that is departing is GONE!

Solutions:
1) if i'm told to remove one unit -- only try to remove that ONE unit -- ignore trying to reap all the others for now
2) If i'm not told which unit to remove, then we should only remove units in the `app_databag` that aren't in the current relation databag.  That's why it's **SO IMPORTANT** we put the `state + node` in the app_databag when we first make a token for someone.  If not, we might make a token, it gets used by a node to join, but it doesn't TELL us that it joined, so we can't remove it.

charms/worker/k8s/src/token_distributor.py
```diff
@@ -305,13 +344,20 @@ def revoke_tokens(
 
         status.add(ops.MaintenanceStatus(f"Revoking {token_type.value} tokens"))
         for unit in remove:
-            if hostname := app_databag.get(unit.name):
+            if node_state := app_databag.get(unit.name):
+                state, node = node_state.split("-", 1)
```
addyess, line 348: whether in the `pending` or `joined` state, we will try to remove that unit

charms/worker/k8s/src/token_distributor.py
```diff
@@ -305,13 +344,20 @@ def revoke_tokens(
...
-                    hostname,
+                    state,
+                    node,
                 )
-                token_strat(hostname, token_type)
+                # ignore pending state, because we can't confirm
+                # if the other node ever used the token to join
+                # ignore self revocation
+                ignore_errors = self.node_name == node or state == "pending"
+                token_strat(node, ignore_errors)
```
addyess, line 361: We can `ignore_errors` on the pending unit.  Here's why -- maybe we gave a token to that unit that it never used b/c it failed to get there.  SO we could hit failures on removing a node that's not in the cluster.  Since there's NO API in the snap to tell me which nodes are in the cluster (???? I mean really ????) We have to keep up with POSSIBLE nodes in the `app_databag`

charms/worker/k8s/src/token_distributor.py
addyess: Fixed

charms/worker/k8s/src/token_distributor.py
```diff
@@ -164,24 +164,24 @@ def _create_cos_token(self, name: str, _):
...
-        except (K8sdConnectionError, InvalidResponseError):
-            if name == self.node_name:
-                # I'm unclustering myself
+        except (K8sdConnectionError, InvalidResponseError) as e:
+            if ignore_errors:
                 # Let's just ignore some of these expected errors:
                 # "Remote end closed connection without response"
                 # "Failed to check if node is control-plane"
-                log.exception("Unclustering ones self can expect an error")
+                log.warning("Remove_Node %s: but with an expected error: %s", name, e)
```
addyess, line 184: Having a stack-trace here wasn't useful debug.

mateoflorido (approved):
Great work! LGTM 👍🏻

mateoflorido:
fixes #85

PR #89, opened 2024-04-30
Added 24.04 base

beliaev-maksim:


addyess:
If i'm not correct, please tell me -- but i've heard noble is  not supported with charmcraft 2.x.   I'd love to be suprised

addyess:
/canonical/self-hosted-runners/run-workflows 1fa151e880a19d6f9b1a66b0f5cbf3b022f38c52

addyess (approved):
I looked at the manifests created and they look sane. I have no idea if this works on juju yet.

beliaev-maksim:
@addyess we are not building on noble but need to execute on noble
Should be irrelevant of charmcraft

addyess:
> @addyess we are not building on noble but need to execute on noble
> Should be irrelevant of charmcraft

Perfect.  Let's go

Issue #75, opened 2024-04-22
worker goes into error after CP departed


beliaev-maksim:
### Bug Description

after https://github.com/canonical/k8s-operator/issues/74 

I just `juju remove-application k8s` and got error state on both workers

```
maksim@darmbeliaev:~$ juju status
Model          Controller              Cloud/Region         Version  SLA          Timestamp
canonical-k8s  k8s-machines-contoller  localhost/localhost  3.4.2    unsupported  17:15:17+02:00

App         Version  Status  Scale  Charm       Channel      Rev  Exposed  Message
k8s-worker  1.30.0   error       2  k8s-worker  latest/edge   34  no       hook failed: "cluster-relation-departed"

Unit           Workload  Agent  Machine  Public address  Ports  Message
k8s-worker/0   error     idle   1        10.112.13.21           hook failed: "cluster-relation-departed"
k8s-worker/1*  error     idle   2        10.112.13.10           hook failed: "cluster-relation-departed"

Machine  State    Address       Inst id              Base          AZ  Message
1        started  10.112.13.21  manual:10.112.13.21  ubuntu@22.04      Manually provisioned machine
2        started  10.112.13.10  manual:10.112.13.10  ubuntu@22.04      Manually provisioned machine
4        started  10.112.13.54  manual:10.112.13.54  ubuntu@22.04      Manually provisioned machine

```

### Relevant log output

```shell
unit-k8s-worker-1: 17:12:59 ERROR unit.k8s-worker/1.juju-log cluster:2: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-k8s-worker-1/charm/venv/ops/model.py", line 3022, in _run
    result = subprocess.run(args, **kwargs)  # type: ignore
  File "/usr/lib/python3.10/subprocess.py", line 526, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '('/var/lib/juju/tools/unit-k8s-worker-1/relation-get', '-r', '2', '-', 'k8s-worker/1', '--format=json')' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-k8s-worker-1/charm/./k8s/src/charm.py", line 596, in <module>
    ops.main.main(K8sCharm)
  File "/var/lib/juju/agents/unit-k8s-worker-1/charm/venv/ops/main.py", line 544, in main
[...]

Issue #74, opened 2024-04-22
Unit in error after teardown


beliaev-maksim:
### Bug Description

Unit goes into unrecoverable error

### To Reproduce

1. deploy CP
2. deploy 2 worker units
3. relate
4. scale CP to 0 units
5. scale CP to 1 unit

### Environment

edge

### Relevant log output

```shell
unit-k8s-1: 17:07:02 ERROR unit.k8s/1.juju-log cluster:1: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-k8s-1/charm/./src/charm.py", line 596, in <module>
    ops.main.main(K8sCharm)
  File "/var/lib/juju/agents/unit-k8s-1/charm/venv/ops/main.py", line 544, in main
    manager.run()
  File "/var/lib/juju/agents/unit-k8s-1/charm/venv/ops/main.py", line 520, in run
    self._emit()
  File "/var/lib/juju/agents/unit-k8s-1/charm/venv/ops/main.py", line 509, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name)
  File "/var/lib/juju/agents/unit-k8s-1/charm/venv/ops/main.py", line 143, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-k8s-1/charm/venv/ops/framework.py", line 352, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-k8s-1/charm/venv/ops/framework.py", line 851, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-k8s-1/charm/venv/ops/framework.py", line 941, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-k8s-1/charm/venv/charms/reconciler.py", line 35, in reconcile
    self.reconcile_function(event)
  File "/var/lib/juju/agents/unit-k8s-1/charm/./src/charm.py", line 476, in _reconcile
    self._revoke_cluster_tokens()
  File "/var/lib/juju/agents/unit-k8s-1/charm/./src/charm.py", line 304, in _revoke_cluster_tokens
    self.distributor.revoke_tokens(
  File "/var/lib/juju/agents/unit-k8s-1/charm/src/token_distributor.py", line 316, in revoke_tokens
    token_strat(hostname, token_type)
[...]

PR #91, opened 2024-05-01
Cluster a node only once, block worker units which are asked to recluster with a new control-plane

addyess:
Address #74 and  #75 

## Overview
Prevent reclustering errors after removing all control-plane units without destroying the control-plane application itself.

## Details
* If all control-plane units are removed  --  a new one would constitute a new cluster
  Therefore existing workers should not try to cluster with the new control-plane units.
* on API join and the node is already clustered, ignore error response
* on API remove and the node is not joined, ignore error response
* dying nodes never reconcile (any juju event will be handled with reconciler)
* synchronize cluster relation using a unique `cluster-name` from the leader control-plane
* synchronize cluster relation using a response from units on the  `joined` field
* ignore errors while removing tokens when:
   1) node is removing itself
   2) token may not have been used (it was a pending token)
   3) local `cluster-name` doesn't match remote `cluster-name`

### DIFF
--- a/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
+++ b/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
@@ -26,6 +26,7 @@ except K8sdAPIManagerError as e:
 Similarly, the module allows for requesting authentication tokens and
 managing K8s components.
 """
+import enum
 import json
 import logging
 import socket
@@ -49,6 +50,18 @@ LIBPATCH = 2
 logger = logging.getLogger(__name__)
 
 
+class ErrorCodes(enum.Enum):
+    """Enumerate the response codes from the k8s api.
+
+    Attributes:
+        StatusNodeUnavailable: returned when the node isn't in the cluster
+        StatusNodeInUse: returned when the node is in the cluster already
+    """
+
+    StatusNodeUnavailable = 520
+    StatusNodeInUse = 521
+
+
 class K8sdAPIManagerError(Exception):
     """Base exception for K8sd API Manager errors."""
 
@@ -58,7 +71,21 @@ class K8sdConnectionError(K8sdAPIManagerError):
 
 
 class InvalidResponseError(K8sdAPIManagerError):
-    """Raised when the response is invalid or unexpected."""
+    """Raised when the response is invalid or unexpected.
+
+    Attributes:
+        code (int): HTTP Status code
+    """
+
+    def __init__(self, code: int, msg: str) -> None:
+        """Initialise the InvalidResponseError.
+
+        Args:
+            code (int): http response code
+            msg (str): Message associated with the error
+        """
+        super().__init__(f"Error status {code}\n" + msg)
+        self.code = code
 
 
 class BaseRequestModel(BaseModel):
@@ -649,17 +676,18 @@ class K8sdAPIManager:
                 data = response.read().decode()
                 if not 200 <= response.status < 300:
                     raise InvalidResponseError(
-                        f"Error status {response.status}\n"
+                        response.status,
                         f"\tmethod={method}\n"
                         f"\tendpoint={endpoint}\n"
                         f"\treason={response.reason}\n"
-                        f"\tbody={data}"
+                        f"\tbody={data}",
                     )
             return response_cls.parse_raw(data)
 
         except ValueError as e:
             raise InvalidResponseError(
-                f"Request failed:\n" f"\tmethod={method}\n" f"\tendpoint={endpoint}"
+                response.status,
+                f"Request failed:\n" f"\tmethod={method}\n" f"\tendpoint={endpoint}",
             ) from e
         except (socket.error, HTTPException) as e:
             raise K8sdConnectionError(
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -126,7 +126,7 @@ class K8sCharm(ops.CharmBase):
         self.reconciler = Reconciler(self, self._reconcile)
         self.distributor = TokenDistributor(self, self.get_node_name(), self.api_manager)
         self.collector = TokenCollector(self, self.get_node_name())
-        self.labeler = LabelMaker(
+        self.labeller = LabelMaker(
             self, kubeconfig_path=self._internal_kubeconfig, kubectl=KUBECTL_PATH
         )
         self._stored.set_default(is_dying=False, cluster_name="")
@@ -696,8 +696,8 @@ class K8sCharm(ops.CharmBase):
         """Apply labels to the node."""
         status.add(ops.MaintenanceStatus("Ensuring Kubernetes Node Labels"))
         node = self.get_node_name()
-        if self.labeler.active_labels() is not None:
-            self.labeler.apply_node_labels()
+        if self.labeller.active_labels() is not None:
+            self.labeller.apply_node_labels()
             log.info("Node %s labelled successfully", node)
         else:
             log.info("Node %s not yet labelled", node)
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -217,7 +217,7 @@ class K8sCharm(ops.CharmBase):
                     self._is_node_present(),
                 ]
             ):
-                self._stored.cluster_name = self.collector.joined_cluster_name(relation)
+                self._stored.cluster_name = self.collector.cluster_name(relation, True)
 
         log.info("%s(%s) current cluster-name=%s", unit, node, self._stored.cluster_name)
         return self._stored.cluster_name
@@ -555,10 +555,9 @@ class K8sCharm(ops.CharmBase):
         self._update_status()
         self._last_gasp()
 
+        relation = self.model.get_relation("cluster")
         local_cluster = self.get_cluster_name()
-        remote_cluster = (
-            relation := self.model.get_relation("cluster")
-        ) and self.collector.recover_cluster_name(relation)
+        remote_cluster = self.collector.cluster_name(relation, False) if relation else ""
         if local_cluster and local_cluster != remote_cluster:
             status.add(ops.BlockedStatus("Cannot rejoin new cluster - remove unit"))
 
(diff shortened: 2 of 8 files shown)


charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -49,6 +50,18 @@ class utilises different connection factories (UnixSocketConnectionFactory
...
+
+    Attributes:
+        StatusNodeUnavailable: returned when the node isn't in the cluster
+        StatusNodeInUse: returned when the node is in the cluster already
+    """
+
+    StatusNodeUnavailable = 520
+    StatusNodeInUse = 521
+
+
```
addyess, line 64: These will come in handy once this is available in the snap
* https://github.com/canonical/k8s-snap/pull/388

they aren't necessary but they are a much better indication of whether a unit is already in the cluster, or not yet in the cluster

charms/worker/k8s/src/charm.py
```diff
@@ -95,14 +96,17 @@ def _cluster_departing_unit(event: ops.EventBase) -> Union[Literal[False], ops.U
     )
 
 
+class NodeRemovedError(Exception):
+    """Raised to prevent reconciliation of dying node."""
+
+
```
addyess, line 102: Once a unit evaluates that it is marked for removal, it should never reconcile afterwards

charms/worker/k8s/src/charm.py
```diff
@@ -125,7 +129,7 @@ def __init__(self, *args):
         self.labeler = LabelMaker(
             self, kubeconfig_path=self._internal_kubeconfig, kubectl=KUBECTL_PATH
         )
-        self._stored.set_default(removing=False)
+        self._stored.set_default(is_dying=False, cluster_name="")
```
addyess, line 132: a unit should keep in stored state the first cluster-name is joins,  It should block if its ever asked to join a different cluster name.

charms/worker/k8s/src/charm.py
```diff
@@ -197,6 +196,32 @@ def _apply_proxy_environment(self):
...
+        Note: It won't change for the lifetime of the unit.
+
+        Returns:
+            the cluster name.
+        """
+        unit, node = self.unit.name, self.get_node_name()
+        if not self._stored.cluster_name:
+            if self.lead_control_plane and self.api_manager.is_cluster_bootstrapped():
+                # TODO: replace with API call once available from the snap
+                self._stored.cluster_name = str(uuid.uuid4())
```
addyess, line 211: Bootstrapped leader control-plane nodes can just make up a cluster name right now.  Something that is unique.  We should replace this by querying the kubernetes cluster itself -- but for now this is okay.

charms/worker/k8s/src/charm.py
```diff
@@ -197,6 +196,32 @@ def _apply_proxy_environment(self):
...
+            elif not (relation := self.model.get_relation("cluster")):
+                pass
```
addyess, line 213: If this unit (worker) isn't clustered -- then pass, there's no node name to set yet

control-planes will always have the cluster relation because its a peer relation

charms/worker/k8s/src/charm.py
```diff
@@ -197,6 +196,32 @@ def _apply_proxy_environment(self):
...
+            elif any(
+                [
+                    self.is_control_plane and self.api_manager.is_cluster_bootstrapped(),
+                    self._is_node_present(),
+                ]
+            ):
```
addyess, line 219: If i'm a control-plane unit that's bootstrapped, or my node is present in the kubernetes cluster, then i'm clustered finally and should set my `cluster_name` based on what's received from the leader control-plane unit who gave me the join token

charms/worker/k8s/src/charm.py
```diff
@@ -489,24 +510,6 @@ def _get_proxy_env(self) -> Dict[str, str]:
...
-        result = subprocess.check_output(shlex.split(cmd))
-        output = result.decode().strip()
-        match = re.search(r"(\d+\.\d+(?:\.\d+)?)", output)
-
-        if match:
-            return match.group()
-
-        log.info("Snap k8s not found or no version available.")
-        return None
-
```
addyess, line 509: this moved into `snap.py`

charms/worker/k8s/src/charm.py
```diff
@@ -519,11 +522,7 @@ def _join_cluster(self):
             status.add(ops.BlockedStatus("Missing cluster integration"))
             assert False, "Missing cluster integration"  # nosec
 
-        if self.is_control_plane and self.api_manager.is_cluster_bootstrapped():
-            relation.data[self.unit]["joined"] = self.get_node_name()
-            return
-
-        if self.collector.joined(relation):
+        if self.get_cluster_name():
```
addyess, line 525: this logic gets pushed into `get_cluster_name`

charms/worker/k8s/src/charm.py
```diff
@@ -541,6 +540,30 @@ def _join_cluster(self):
...
+        """
+        if self.lead_control_plane:
+            self._revoke_cluster_tokens(event)
+        self._update_status()
+        self._last_gasp()
+
+        local_cluster = self.get_cluster_name()
+        remote_cluster = (
+            relation := self.model.get_relation("cluster")
+        ) and self.collector.recover_cluster_name(relation)
```
addyess, line 561: black thinks this is readable????

charms/worker/k8s/src/charm.py
```diff
@@ -541,6 +540,30 @@ def _join_cluster(self):
...
+        if local_cluster and local_cluster != remote_cluster:
+            status.add(ops.BlockedStatus("Cannot rejoin new cluster - remove unit"))
+
+        raise NodeRemovedError()
+
```
addyess, line 566: If we're dying -- never reconcile.

charms/worker/k8s/src/charm.py
```diff
@@ -577,51 +595,76 @@ def _reconcile(self, event: ops.EventBase):
...
-        AssertionError,
-        subprocess.CalledProcessError,
-        InvalidResponseError,
-        K8sdConnectionError,
-    )
     def _update_status(self):
         """Check k8s snap status."""
-        if self.is_dying:
-            status.add(ops.WaitingStatus("Preparing to leave cluster"))
-            return
```
addyess, line 591: the `death_handler` sets to `WaitingStatus("Awaiting cluster removal")`.  This wasn't necessary anymore

charms/worker/k8s/src/charm.py
```diff
@@ -577,51 +595,76 @@ def _reconcile(self, event: ops.EventBase):
...
-        if self.is_control_plane:
-            assert self.api_manager.is_cluster_ready(), "control-plane not yet ready"  # nosec
```
addyess, line 593: a better check from `self.get_cluster_name()` tells us if this node is clustered or not

charms/worker/k8s/src/charm.py
```diff
@@ -577,51 +595,76 @@ def _reconcile(self, event: ops.EventBase):
...
-
-        if version := self._get_snap_version():
+        if version := snap_version("k8s"):
             self.unit.set_workload_version(version)
 
+        if not self.get_cluster_name():
+            status.add(ops.WaitingStatus("Node not Clustered"))
+            return
```
addyess, line 605: If not bootstrapped, or waiting to join a cluster, you get this Waiting Status

charms/worker/k8s/src/charm.py
```diff
@@ -577,51 +595,76 @@ def _reconcile(self, event: ops.EventBase):
...
+
         if not self._is_node_ready():
-            status.add(ops.WaitingStatus("Node not yet Ready"))
+            status.add(ops.WaitingStatus("Node not Ready"))
             return
```
addyess, line 609: If the node is IN the cluster, but has a `NotReady` status (or it can't be found b/c the api server is down), they'll see this

charms/worker/k8s/src/charm.py
```diff
@@ -577,51 +595,76 @@ def _reconcile(self, event: ops.EventBase):
...
             event: ops.EventBase - event that triggered charm hook
+
+        Returns:
+            True if being removed, otherwise False
         """
-        if self.is_dying:
-            return
-        if unit := _cluster_departing_unit(event):
+        if self._stored.is_dying:
+            pass
```
addyess, line 621: I know i'm dying -- skip other checks

charms/worker/k8s/src/charm.py
```diff
@@ -577,51 +595,76 @@ def _reconcile(self, event: ops.EventBase):
...
+        elif (unit := _cluster_departing_unit(event)) and unit == self.unit:
             # Juju says I am being removed
-            self._stored.removing = unit == self.unit
```
addyess, line 612: this adjustment let's us keep evaluating the `elif` chain on relation-departed hooks.   This comes in handy when workers handled a relation-departed hook because the last control-plane unit is departing the relation.  the departing unit is the `k8s/N` -- but since its the last one, the k8s-worker needs to die as well

charms/worker/k8s/src/charm.py
```diff
@@ -577,51 +595,76 @@ def _reconcile(self, event: ops.EventBase):
...
-        elif isinstance(event, ops.RelationBrokenEvent) and event.relation.name == "cluster":
+            self._stored.is_dying = True
+        elif (
+            isinstance(event, ops.RelationBrokenEvent)
+            and event.relation.name == "cluster"
+            and self.is_worker
+        ):
```
addyess, line 629: again, this check changes a bit to help keep processing the `elif` chain

tests/integration/test_k8s.py
```diff
@@ -58,10 +58,10 @@ async def test_nodes_labelled(request, kubernetes_cluster: model.Model):
         juju_nodes = [n for n in nodes if "juju-charm" in n["metadata"]["labels"]]
         assert len(k8s.units + worker.units) == len(
             labelled
-        ), "Not all nodes labeled with custom-label"
+        ), "Not all nodes labelled with custom-label"
```
addyess, line 61: fixed spelling

mateoflorido (approved):
Amazing work! LGTM

PR #98, opened 2024-05-03
Initial version of External Certificates integration

mateoflorido:
#### Applicable spec
KU083 - Vault integration

## Overview
Implement the External Certificates relation using TLSCertificatesV3.

### Rationale
This pull request introduces the capability of integrating the charm with external certificate providers like Vault or charms that use the TLSCertificatesV3 library, such as `self-signed-certificates`.

This pull request only tackles the bootstrap certificates generation. The control plane and worker certificates will be implemented in separate pull requests in the future.

### Module Changes
- Added a K8sCertificates module, which manages the request process for the Kubernetes' components certificates.

### Library Changes
- Added TLSCertificatesV3.


addyess:
/canonical/self-hosted-runners/run-workflows 14a739b

charms/worker/k8s/charmcraft.yaml
```diff
@@ -98,7 +106,16 @@ actions:
...
+    charm-binary-python-packages:
+      - setuptools
+      - cryptography
+    build-packages:
+      - git
+      - libffi-dev
+      - libssl-dev
+      - pkg-config
+      - rustc
+      - cargo
```
addyess, line 118: Is there some chance these packages which are used by the focal build machine create something that doesn't work in jammy/noble.  Hopefully not.  I know our integration tests only run on focal ATM

Perhaps we need a matrix of testing against jammy/noble in the future

charms/worker/k8s/src/certificates.py
```diff
@@ -0,0 +1,280 @@
...
+            name = csr.replace("-csr", "")
+            if self._check_secret(label=f"{name}-cert-{self.charm.unit.name}"):
+                certificate_secret = self.charm.model.get_secret(
+                    label=f"{name}-cert-{self.charm.unit.name}"
+                )
+                certificate_secret.set_content(content=certificate_content)
+            else:
+                self.charm.unit.add_secret(
+                    content=certificate_content, label=f"{name}-cert-{self.charm.unit.name}"
+                )
```
addyess, line 192: Surely this certificate expires?  Do you need to read the expiry on the certificate and expire the secret at the same time.  How do you handle regenerating secrets?  Is there someway to test the certificate rotation by changing the expiration time to something super short?

charms/worker/k8s/src/certificates.py
```diff
@@ -0,0 +1,280 @@
...
+        Args:
+            components (List[str]): A list of components to generate private keys for.
+        """
+        content = dict()
+
+        for key in components:
+            private_key = generate_private_key()
+            content[key] = private_key.decode()
+
+        self.charm.unit.add_secret(content=content, label=f"{self.charm.unit.name}-private-keys")
```
addyess, line 64: Does it make sense that these csr/private-keys would ever need rotation?  Should they have secrets that expire and need to be regenerated?

addyess (commented):
I have questions concerning the expiration of certificates, testing of those situations, and testing of other ubuntu bases.  I don't expect we address all of them at once.  

the most worrisome to me is certificate expiration

charms/worker/k8s/requirements.txt
```diff
@@ -1,7 +1,10 @@
+backports.cached-property==1.0.2
```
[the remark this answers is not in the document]
addyess: yeah, i don't understand why we'd need that.  
[functools.cached_property](https://docs.python.org/3.8/library/functools.html#functools.cached_property) was a thing in python 3.8

charms/worker/k8s/src/certificates.py
mateoflorido: That's correct, the library emits events when the certificates are near to the expiry date. The certificate expiration should be implemented in another PR. We still need to implement the mechanism into the snap to allow updating the certificate bits in the cluster. I've created KU-776 for this purpose.

charms/worker/k8s/src/certificates.py
mateoflorido: I don't think so. When we receive an event that signals an expiring certificate, we need to provide both the old CSR and the private key to attempt the renewal. For instance:
```python
    def _on_certificate_expiring(
        self, event: Union[CertificateExpiringEvent, CertificateInvalidatedEvent]
    ) -> None:
        replicas_relation = self.model.get_relation("replicas")
        if not replicas_relation:
            self.unit.status = WaitingStatus("Waiting for peer relation to be created")
            event.defer()
            return
        old_csr = replicas_relation.data[self.app].get("csr")
        private_key_password = replicas_relation.data[self.app].get("private_key_password")
        private_key = replicas_relation.data[self.app].get("private_key")
        new_csr = generate_csr(
            private_key=private_key.encode(),
            private_key_password=private_key_password.encode(),
            subject=self.cert_subject,
        )
        self.certificates.request_certificate_renewal(
            old_certificate_signing_request=old_csr,
            new_certificate_signing_request=new_csr,
        )
        replicas_relation.data[self.app].update({"csr": new_csr.decode()})
```
Seeing this usage from the library examples, I think we need to keep both CSR and Private Keys to ease the certificate management.

charms/worker/k8s/charmcraft.yaml
mateoflorido: Sure, I'll create a card to track this process

charms/worker/k8s/src/charm.py
```diff
@@ -310,7 +304,7 @@ def _configure_datastore(self, config: Union[BootstrapConfig, UpdateClusterConfi
         """Configure the datastore for the Kubernetes cluster.
 
         Args:
-            config (BootstrapConfig|UpdateClusterConfigRequst):
+            config (BootstrapConfig|UpdateClusterConfigRequest):
```
addyess, line 307: 😃

tests/integration/conftest.py
```diff
@@ -170,9 +167,13 @@ def switch(self, name: str, path: Path):
...
-        app = self.applications[name]
-        app["charm"] = str(path.resolve())
-        app["channel"] = None
+        # FIXME: Omit non present charms
+        try:
+            app = self.applications[name]
+            app["charm"] = str(path.resolve())
+            app["channel"] = None
+        except KeyError:
+            log.warning("Application %s not found in bundle", name)
```
addyess, line 176: whats this for? oh? for test bundles where theres no worker charm?

addyess (approved):
amazing 😍 Lgtm

tests/integration/conftest.py
mateoflorido: Yes, it should be there until we add support for certificates in k8s-worker

PR #99, opened 2024-05-06
Pin snap revision for amd64

addyess:
## Summary
Pins the snap revision rather than using a snap channel

## Changes
Doing so, I discovered an issue with the logic within the snap library.  Every time you call `snap.ensure` it will refresh the snap, even if the snap revisions match.  This caused some of the services of k8s to be stopped, and never restarted. 

By checking the current revision against the target snap revision, we can prevent the charm from calling `snap refresh` over and over again


bschimke95 (approved):
LGTM

mateoflorido (approved):
LGTM2

PR #101, opened 2024-05-06
pin operator-workflow to a commit (#100)

addyess:
* pin operator-workflow to a commit
* No long raise_on_blocked, worker charms will be blocked by default until the unit processes the join cluster relation hook


addyess:
Maybe we can get away with not using `squash` and merge on this one to keep the commit trees the same

mateoflorido (approved):

PR #105, opened 2024-05-08
Refactor integration tests to be re-entrant

addyess:
### Overview
This effort is to make the integration-tests runnable against an existing  model, have the test detect the correct model is deployed, and run the tests in a way that is re-entrant (ie -- the tests can be run over and over without disturbing the existing model)

### Rationale
SolutionsQA has a request to deploy the Release Candidate bundle on many clouds types, then test that deployed model.  This would also aid in development of tests because an existing model could be reused to develop tests against.

### Details
* detection of existing installation of canonical-kubernetes in a model
* all etcd tests pass with 1 or 3 etcd units
* all k8s tests pass even with etcd datastore backend
* applies proxy config to model where squid.internal with command line switch `--apply-proxy`

example:
```bash
tox -e integration -- --model test --apply-proxy -k test_k8s 
```

### Checklist
- [x] The documentation is generated using `src-docs`
- [x] The documentation for charmhub is updated
- [x] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)


tests/integration/conftest.py
```diff
@@ -192,6 +192,40 @@ async def cloud_type(ops_test: OpsTest):
...
+        ops_test (OpsTest): ops_test plugin
+
+    Returns:
+        true if proxy was configured.
+    """
+    assert ops_test.model, "Model must be present"
+    controller = await ops_test.model.get_controller()
+    controller_model = await controller.get_model("controller")
+    controller_unit = controller_model.applications["controller"].units[0]
+    http_proxy, site = "http://squid.internal:3128", "https://ghcr.io"
```
bschimke95, line 211: If Github is down, this code will think that this is not run behind a proxy. Maybe add another site e.g. google to double-check in case Github is down (again).

tests/integration/data/static-no-proxy.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+- "localhost"
+- "127.0.0.1"
+- "0.0.0.0"
+- "ppa.launchpad.net"
+- "launchpad.net"
+- "10.0.8.0/24"
+- "10.152.183.0/24"
```
bschimke95, line 10: nit: newline

tests/integration/conftest.py
```diff
@@ -253,18 +287,27 @@ async def kubernetes_cluster(request: pytest.FixtureRequest, ops_test: OpsTest):
     bundle_marker = request.node.get_closest_marker("bundle_file")
     if bundle_marker:
         bundle_file = bundle_marker.args[0]
+    bundle_path = Path(__file__).parent / "data" / bundle_file
+    model = "main"
+
+    with ops_test.model_context(model) as the_model:
+        if await is_deployed(the_model, bundle_path):
```
bschimke95, line 294: Will this fail if the tests failed in the clustering tests?
e.g. https://github.com/canonical/k8s-operator/blob/main/tests/integration/test_k8s.py#L104

Let's say the leader was removed [here](https://github.com/canonical/k8s-operator/blob/main/tests/integration/test_k8s.py#L108) but the next step failed. If the tests are rerun, the `is_deployed` check will fail because the number of expected units is wrong and the bundle is freshly deployed, right?

bschimke95 (commented):
Looks great. Left some comments for my understanding.

tests/integration/conftest.py
addyess: RIght... this isn't the STRONGEST of tests to determine if there's a proxy config necessary.  I thought about just having a pytest command line switch that says "apply proxy config" -- maybe that's a safer option than trying to TEST if a site is available from behind some proxy. 

WDYT?

tests/integration/conftest.py
addyess: That's mostly correct.  If the `is_deployed` check fails, it will apply the bundle to the model as if the model was empty

This fixture `kubernetes_cluster` has `module` scope meaning that it's run once for each python test module.  In the case of test_k8s -- if the fixture completes successfully once, then each test uses the model that fixture yields.  this `is_deployed` won't be checked between tests runs within that file. 

But if say the tests FAIL, and pytest ends, then a user re-runs the tests again -- this fixture would catch there were not enough `k8s` units, and deploy again from the bundle which should yield the right number of k8s units.

bschimke95: makes sense - thanks!

bschimke95 (approved):
LGTM

tests/integration/conftest.py
bschimke95: This sounds safer, indeed

Issue #107, opened 2024-05-09
ERROR os "ubuntu" version "24.04" not found


mcarvalhor:
### Bug Description

I couldn't deploy Canonical K8s charm on an series other than 24.04 - even when explicitly specifying the Ubuntu series.

If I run:
`juju deploy k8s my-k8s-test-application --channel=latest/edge --revision=59 --series=jammy`
Juju says:
`WARNING series flag is deprecated, use --base instead
ERROR os "ubuntu" version "24.04" not found`

If I run:
`juju deploy k8s my-k8s-test-application --channel=latest/edge --revision=59 --base='ubuntu@24.04'`
Juju says:
`ERROR os "ubuntu" version "24.04" not found`

The current workaround is: using revision 47 - later revisions fail.

### To Reproduce

1. juju deploy k8s my-k8s-test-application --channel=latest/edge --revision=59 --base='ubuntu@22.04'
2. Fails.

### Environment

stg-is-new-canonical-k8s-testing@is-bastion-ps6

Channel: latest/edge
Failing revisions: 48 and above
Working revisions: 47 and below
Juju Controller/Model: 3.1.8
Juju client: 3.1.8-ubuntu-amd64

### Relevant log output

```shell
`juju debug-log` is empty because nothing is even deployed.
```

### Additional context

We have Ubuntu 24.04 available on the cloud.
For instance, `juju add-machine --base="ubuntu@24.04"` works and creates a Noble VM.

addyess:
https://discourse.charmhub.io/t/juju-support-for-the-latest-lts-release-noble-numbat-24-04/14051

i'm not sure how to deploy to noble.  I think this is a issue of juju

```
⚡ juju deploy k8s focal-k8s --channel=1.30/edge --base=ubuntu@20.04
Deployed "focal-k8s" from charm-hub charm "k8s", revision 58 in channel 1.30/beta on ubuntu@20.04/stable

addyess@addyess-dell /home/addyess                                                                                           
⚡ juju deploy k8s jammy-k8s --channel=1.30/edge --base=ubuntu@22.04
Deployed "jammy-k8s" from charm-hub charm "k8s", revision 58 in channel 1.30/beta on ubuntu@22.04/stable

addyess@addyess-dell /home/addyess                                                                                           
⚡ juju deploy k8s noble-k8s --channel=1.30/edge --base=ubuntu@24.04
ERROR os "ubuntu" version "24.04" not found
```

mcarvalhor:
[...]

PR #113, opened 2024-05-13
`runs-on` noble requires at least juju3.3 and beyond

addyess:
Resolves https://github.com/canonical/k8s-operator/issues/107

It seems that a bug in juju 3.1 prevents the k8s charm from deploying on any series (jammy or focal) because it mentions an unsupported series (noble)

The message from juju is cryptic (see issue) so to not confuse users, it seems most prudent to just update the minimum version of the charm to 3.3\

Will need to cherry-pick to main on merge


bschimke95 (approved):
LGTM

PR #116, opened 2024-05-17
Prefer the cluster relation for bootstrap address rather than juju-info

addyess:
Swap from `juju-info` relation to `cluster` relation when determining which address to bind for bootstrapping


mateoflorido (approved):
LGTM

PR #119, opened 2024-05-22
enable ceph-csi integration

kwmonroe:
### Overview

Enable Ceph CSI integration.

### Rationale

We're recommending ceph as the storage solution when using Canonical K8s as the backing cloud for `juju add-k8s`. This PR provides the integration needed by the `ceph-csi` subordinate charm.

### Juju Events Changes

Adds a new relation-changed observation for `ceph-k8s-info`.

### DIFF
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -153,11 +153,7 @@ class K8sCharm(ops.CharmBase):
             self.etcd = EtcdReactiveRequires(self)
             self.framework.observe(self.on.get_kubeconfig_action, self._get_external_kubeconfig)
 
-        # kubernetes-info
-        if self.lead_control_plane:
-            self.framework.observe(self.on.ceph_k8s_info_relation_changed, self._k8s_info)
-
-    def _k8s_info(self, event: ops.RelationChangedEvent):
+    def _k8s_info(self, event: ops.EventBase):
         """Send cluster information on the kubernetes-info relation.
 
         Provide applications with cluster characteristics. This should only run on the lead
@@ -166,12 +162,8 @@ class K8sCharm(ops.CharmBase):
         Args:
             event: ops.RelationChangedEvent - event triggered by the relation changed hook
         """
-        if not self.lead_control_plane:
-            return
-
-        if event.relation.name == "ceph-k8s-info":
-            rel = event.relation
-            rel.data[self.app]["kubelet-root-dir"] = "/var/lib/kubelet"
+        if isinstance(event, ops.RelationChangedEvent) and event.relation.name == "ceph-k8s-info":
+            event.relation.data[self.app]["kubelet-root-dir"] = "/var/lib/kubelet"
 
     @status.on_error(
         ops.WaitingStatus("Installing COS requirements"),
@@ -699,6 +691,7 @@ class K8sCharm(ops.CharmBase):
         self._apply_snap_requirements()
         self._check_k8sd_ready()
         if self.lead_control_plane:
+            self._k8s_info(event)
             self._bootstrap_k8s_snap()
             self._enable_functionalities()
             self._update_annotations()
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -153,6 +153,26 @@ class K8sCharm(ops.CharmBase):
             self.etcd = EtcdReactiveRequires(self)
             self.framework.observe(self.on.get_kubeconfig_action, self._get_external_kubeconfig)
 
+        # kubernetes-info
+        if self.lead_control_plane:
+            self.framework.observe(self.on.ceph_k8s_info_relation_changed, self._k8s_info)
+
+    def _k8s_info(self, event: ops.RelationChangedEvent):
+        """Send cluster information on the kubernetes-info relation.
+
+        Provide applications with cluster characteristics. This should only run on the lead
+        k8s control plane unit.
+
+        Args:
+            event: ops.RelationChangedEvent - event triggered by the relation changed hook
+        """
+        if not self.lead_control_plane:
+            return
+
+        if event.relation.name == "ceph-k8s-info":
+            rel = event.relation
+            rel.data[self.app]["kubelet-root-dir"] = "/var/lib/kubelet"
+
     @status.on_error(
         ops.WaitingStatus("Installing COS requirements"),
         subprocess.CalledProcessError,
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -184,7 +184,7 @@ class Bundle:
     def drop_constraints(self):
         """Remove constraints on applications. Useful for testing on lxd."""
         for app in self.applications.values():
-            app["constraints"] = None
+            app["constraints"] = ""
 
     def add_constraints(self, constraints: Dict[str, str]):
         """Add constraints to applications.
@@ -193,7 +193,10 @@ class Bundle:
             constraints:  Mapping of constraints to add to applications.
         """
         for app in self.applications.values():
-            val: str = app["constraints"]
+            if app.get("num_units", 0) < 1:
+                log.info("Skipping constraints for subordinate charm: %s", app["charm"])
+                continue
+            val: str = app.get("constraints", "")
             existing = dict(kv.split("=", 1) for kv in val.split())
             existing.update(constraints)
             app["constraints"] = " ".join(f"{k}={v}" for k, v in existing.items())
--- a/tests/integration/data/test-bundle-ceph.yaml
+++ b/tests/integration/data/test-bundle-ceph.yaml
@@ -1,0 +1,44 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: integration-test-ceph
+description: |-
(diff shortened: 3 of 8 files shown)


charms/worker/k8s/src/charm.py
```diff
@@ -149,6 +149,26 @@ def __init__(self, *args):
...
+    def _k8s_info(self, event: ops.RelationChangedEvent):
+        """Send cluster information on the kubernetes-info relation.
+
+        Provide applications with cluster characteristics. This should only run on the lead
+        k8s control plane unit.
+
+        Args:
+            event: ops.RelationChangedEvent - event triggered by the relation changed hook
+        """
+        if not self.lead_control_plane:
```
bschimke95, line 165: This feels like a unnecessary double-check. 

I'd say either remove the outer or inner call since we do control the calling of this private function.

charms/worker/k8s/src/charm.py
```diff
@@ -149,6 +149,26 @@ def __init__(self, *args):
...
+            return
+
+        if event.relation.name == "ceph-k8s-info":
+            rel = event.relation
```
bschimke95, line 169: nit: you are not using this var anywhere else. 
`event.relation.data[self.app][...]` looks fine to me. No need for this extra assignment.

tests/integration/test_ceph.py
```diff
@@ -0,0 +1,38 @@
...
+# This pytest mark configures the test environment to use the Canonical Kubernetes
+# bundle with ceph, for all the test within this module.
+pytestmark = [
+    pytest.mark.bundle_file("test-bundle-ceph.yaml"),
+    pytest.mark.ignore_blocked,
+]
+
+
+@pytest.mark.abort_on_fail
+async def test_nodes_ready(kubernetes_cluster: model.Model):
```
bschimke95, line 23: This does not seem ceph related but rather like a smoke test. Do we need this here?

bschimke95 (commented):
Amazing that we should have to connect the relation and ceph "just works". 
I added some general comments and leave the approval to the charming experts.

charms/worker/k8s/src/charm.py
```diff
@@ -149,6 +149,26 @@ def __init__(self, *args):
             self.etcd = EtcdReactiveRequires(self)
             self.framework.observe(self.on.get_kubeconfig_action, self._get_external_kubeconfig)
 
+        # kubernetes-info
+        if self.lead_control_plane:
+            self.framework.observe(self.on.ceph_k8s_info_relation_changed, self._k8s_info)
```
addyess, line 154: aight, so this doesn't really jive with the `reconcile` pattern presented in this charm.  by default, the control-plane charm will handle `ceph-k8s-info relation-changed` events through the `_reconcile(...)` method without doing this framework observe.  By registering this observe here: the charm is now handling with two methods:
`_reconcile` and `_k8s_info`

I'd recommend every control-plane unit (not just the leader) registers this event just as the reconciler will do
```suggestion
        if self.is_control_plane:
            self.framework.observe(self.on.ceph_k8s_info_relation_changed, self._k8s_info)
```

or maybe just do away with this entirely and put the `self._k8s_info` call into the `_reconcile(...)` method

charms/worker/k8s/src/charm.py
addyess: agree... likely unnecessary depending on the caller.  If it's called from the reconcile function this isn't necessary

charms/worker/k8s/src/charm.py
addyess: ahh, ok, so you're letting the ceph application know where kubelet is.  Very good!  Yes, i'm feeling more and more confident we should follow the reconciler pattern here.

charms/worker/k8s/src/charm.py
addyess: I suggest making the call to `self._k8s_info()` [here](https://github.com/canonical/k8s-operator/pull/119/files#diff-5ba1576e3daac57729756610a1f51b9706129e66f250d7a8557ea5666d0c490aR602)

tests/integration/test_ceph.py
addyess: Agree, not necessary.  I think the second test is more appropriate

tests/integration/test_ceph.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@pytest.mark.abort_on_fail
+async def test_ceph_sc(kubernetes_cluster: model.Model):
+    """Test that a ceph storage class is available."""
+    k8s: unit.Unit = kubernetes_cluster.applications["k8s"].units[0]
+    event = await k8s.run("k8s kubectl get sc -o=jsonpath='{.items[*].provisioner}'")
+    result = await event.wait()
+    stdout = result.results["stdout"]
+    assert "rbd.csi.ceph.com" in stdout, f"No ceph provisioner found in: {stdout}"
```
addyess, line 38: It would be REALLY nice if we had a test that created a PV , creating a pod that uses it, wrote something to it, create another pod that uses that PV, reads the data, and confirms what was written is what was read. 

Maybe a follow up PR?

addyess (changes requested):
Looking really good.  My one ask is to use the reconciler

tests/integration/data/test-bundle-ceph.yaml
```diff
@@ -35,8 +35,12 @@ applications:
...
     num_units: 1
     storage:
-      osd-devices: 1G,2
-      osd-journals: 1G,1
+      osd-devices:
+        count: 2
+        size: '1G'
+      osd-journals:
+        count: 1
+        size: '1G'
```
addyess, line 43: This was is hard-won... but 

[ApplicationDeploy.storage](https://github.com/juju/python-libjuju/blob/master/juju/client/_definitions.py#L1592)
uses [`from_json`](https://github.com/juju/python-libjuju/blob/master/juju/client/facade.py#L663)

where the right side of each key (eg `1G,2`) must either be valid json, or just a `dict` with the correct keys defined in [Constraints](https://github.com/juju/python-libjuju/blob/master/juju/client/_definitions.py#L6511)

addyess: it seems there's very little to be done here with pylibjuju.  If I go with any deviations to the bundle (like this one above), they fail when python-libjuju uses `fetch_plan` to validate the bundle.

I'm not sure how to proceed with storage constraints from pylibjuju

```yaml
    storage:
      osd-devices: '{count: 2, size: "1G"}'
      osd-journals: '{count: 1, size: "1G"}'
```

fails like this:

```
juju.errors.JujuError: ['invalid storage "osd-devices" in application "ceph-osd": unrecognized storage constraint "{count: 2" not valid', 'invalid storage "osd-journals" in application "ceph-osd": unrecognized storage constraint "{count: 1" not valid']
```

addyess: the [pylibjuju example](https://github.com/juju/python-libjuju/blob/53cd33d884f3ca710e1fa026c74f31b02281be69/tests/integration/bundle/bundle-with-storage-constraint.yaml) has the original submission -- with a slight deviation

addyess: but yeah... even the example constraints fails

@cderici take a look here https://github.com/canonical/k8s-operator/actions/runs/9411181657/job/25924049004#step:22:71

cderici: @addyess yeah this appears like a bug in pylibjuju, and quick look shows that we might have more issues there like resolving things like `ceph-osd` as well, can you open a bug on libjuju so I'll have someone take a closer look?

addyess: @cderici I raised this issue https://github.com/juju/python-libjuju/issues/1060 in python-libjuju

addyess (approved):
Waiting now on tests to pass

addyess:
@mateoflorido can you take one final look here?

tests/integration/conftest.py
```diff
@@ -193,7 +193,10 @@ def add_constraints(self, constraints: Dict[str, str]):
             constraints:  Mapping of constraints to add to applications.
         """
         for app in self.applications.values():
-            val: str = app["constraints"]
+            if app.get("num_units", 0) < 1:
+                log.info("Skipping constraints for subordinate charm: %s", app["charm"])
+                continue
```
addyess, line 198: prevent subordinate applications in our bundles `tests/data/*.yaml` from getting constraints added to them

tests/integration/data/test-bundle-ceph.yaml
```diff
@@ -0,0 +1,44 @@
...
+  k8s-worker:
+    charm: k8s-worker
+    channel: latest/edge
+    constraints: cores=2 mem=8G root-disk=16G
+    num_units: 1
+  ceph-csi:
+    charm: ceph-csi
+    channel: latest/stable
+    options:
+      provisioner-replicas: 1
```
addyess, line 23: this is a subordinate charm and it cannot ever have constraints attached.  Notice that `num_units` isn't set.

mateoflorido (approved):
LGTM! Could you please create a follow up card to add more robust integration testing around this feature?

addyess:
> LGTM! Could you please create a follow up card to add more robust integration testing around this feature?

Done

PR #126, opened 2024-05-27
Add Joining Certificates for non-leader Control Planes and Workers

mateoflorido:
### Overview
This pull request adds support for external certificates for non-leader control planes and worker units.

### Rationale
A mechanism is required to allow the joining units to request their own certificates and pass them to the joining configuration object. In this pull request, joining units can request their required certificates over the TLSCertificatesv3 interface, which can be used with charms that implement this interface (e.g., new Vault or self-signed-certificates).

### DIFF
--- a/charms/worker/charmcraft.yaml
+++ b/charms/worker/charmcraft.yaml
@@ -45,6 +45,14 @@ bases:
       architectures: [amd64]
 config:
   options:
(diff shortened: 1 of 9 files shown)


addyess (approved):
This is really coming along.  Thanks @mateoflorido

PR #133, opened 2024-07-15
Address fan-network disabling on aws

addyess:
### Overview

Applies workaround for testing cilium on AWS from [LP#2016905](https://bugs.launchpad.net/charm-cilium/+bug/2016905/comments/1)

### Details

Integration test attempts to determine which cloud the controller is on, in order to apply correct model configuration


mateoflorido (approved):
LGTM

PR #141, opened 2024-08-09
[COS] Update to pydantic 2

addyess:
### Overview

* migrates pydantic 2 structures and validators
* removes pydantic 2 deprecation warnings
* updates cos_agent libraries

### Details
* Update the cos_agent allows us to upgrade to `pydantic ~= 2.*`
* Improve the snap and containerd modules to use pydantic 2 features.
* Update the cosl library to current


tests/integration/conftest.py
```diff
@@ -314,7 +315,16 @@ async def kubernetes_cluster(request: pytest.FixtureRequest, ops_test: OpsTest):
 @pytest_asyncio.fixture(name="_grafana_agent", scope="module")
 async def grafana_agent(kubernetes_cluster: Model):
     """Deploy Grafana Agent."""
-    await kubernetes_cluster.deploy("grafana-agent", channel="stable")
+    primary = kubernetes_cluster.applications["k8s"]
```
bschimke95, line 318: Is this reqiured because the of the updated `cos_agent`?

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -124,24 +126,22 @@ def check_status_code(cls, v):
             raise ValueError(f"Status code must be 200. Received {v}")
         return v
 
-    @validator("error_code", always=True)
-    def check_error_code(cls, v, values):
+    @model_validator(mode="after")
+    def check_error_code(self) -> Self:
```
bschimke95, line 130: I would not expect `check_error_code` to return `self` but I guess this is for chaining checks? Is this a common pattern for charms?

bschimke95 (approved):
LGTM, only questions

bschimke95:
Also, seems like the integration tests are not happy for some reason.

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
addyess: this is a pattern from pydantic validators
https://docs.pydantic.dev/2.8/concepts/validators/#model-validators

tests/integration/conftest.py
```diff
@@ -388,9 +401,13 @@ async def cos_lite_installed(ops_test: OpsTest, cos_model: Model):
...
-async def traefik_address(cos_model: Model, _cos_lite_installed):
-    """Fixture to get Traefik address."""
-    yield await get_address(model=cos_model, app_name="traefik")
+async def traefik_url(cos_model: Model, _cos_lite_installed):
+    """Fixture to fetch Traefik url."""
+    action = await cos_model.applications["traefik"].units[0].run_action("show-proxied-endpoints")
+    action = await action.wait()
+    p_e = json.loads(action.results["proxied-endpoints"])
+
+    yield p_e["traefik"]["url"]
```
addyess, line 410: Uses the charm's `show-proxied-endpoints` action as described [here](https://charmhub.io/topics/canonical-observability-stack/tutorials/install-microk8s#heading--browse-dashboards)

charms/worker/k8s/charmcraft.yaml
```diff
@@ -145,7 +145,11 @@ actions:
 
 parts:
   charm:
-    build-packages: [git]
+    plugin: charm
+    build-packages:
+    - git
+    - rustc
+    - cargo
```
addyess, line 152: yup -- we need `rustc` and `cargo` to build libraries for pydantic... we need to confirm the charms built by focal work on jammy + noble

addyess: Well, this firmly breaks jammy and noble installs because pydantic 2 gets compiled into a binary which can only be used by python38.

```
root@juju-fc684e-0:/var/lib/juju/agents/unit-k8s-0/charm/venv/pydantic_core# file _pydantic_core.cpython-38-x86_64-linux-gnu.so
_pydantic_core.cpython-38-x86_64-linux-gnu.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=c09ad7f4f2d18c122a214f4c169330304771d064, stripped
```

PR #142, opened 2024-08-09
Handle various microcluster ready endpoints

addyess:
### Overview
With the advent of v2 microcluster, the ready endpoint moved from `/cluster/1.0/ready` to `/core/1.0/ready` because changing endpoints is a totally natural thing to do for run

### Details
* loop through available endpoints til we find one that doesn't `404`.  
* reraise any non-404 excptions
* raise a `K8sdConnectionError` if we run out of endpoints to try

### DIFF
--- a/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
+++ b/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
@@ -782,15 +782,17 @@ class K8sdAPIManager:
             K8sdConnectionError: If the response is Not Found on all endpoints.
         """
         ready_endpoints = ["/core/1.0/ready", "/cluster/1.0/ready"]
-        for endpoint in ready_endpoints:
+        for i, endpoint in enumerate(ready_endpoints):
             try:
                 self._send_request(endpoint, "GET", EmptyResponse)
                 break
             except InvalidResponseError as ex:
                 if ex.code == 404:
                     logger.warning(
-                        "Encountered 404 while checking if micro-cluster is ready @ %s: %s",
+                        "micro-cluster unavailable @ %s (%s of %s): %s",
                         endpoint,
+                        i + 1,
+                        len(ready_endpoints),
                         ex,
                     )
                     # Try the next endpoint if the current one is not found
--- a/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
+++ b/charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
@@ -776,9 +776,30 @@ class K8sdAPIManager:
         return status.metadata and status.metadata.status.ready
 
     def check_k8sd_ready(self):
-        """Check if k8sd is ready."""
-        endpoint = "/cluster/1.0/ready"
-        self._send_request(endpoint, "GET", EmptyResponse)
+        """Check if k8sd is ready using various microcluster endpoints.
+
+        Raises:
+            K8sdConnectionError: If the response is Not Found on all endpoints.
+        """
+        ready_endpoints = ["/core/1.0/ready", "/cluster/1.0/ready"]
+        for endpoint in ready_endpoints:
+            try:
+                self._send_request(endpoint, "GET", EmptyResponse)
+                break
+            except InvalidResponseError as ex:
+                if ex.code == 404:
+                    logger.warning(
+                        "Encountered 404 while checking if micro-cluster is ready @ %s: %s",
+                        endpoint,
+                        ex,
+                    )
+                    # Try the next endpoint if the current one is not found
+                    continue
+                raise
+        else:
+            raise K8sdConnectionError(
+                "Exhausted all endpoints while checking if micro-cluster is ready"
+            )
 
     def bootstrap_k8s_snap(self, request: CreateClusterRequest) -> None:
         """Bootstrap the k8s cluster.
--- a/charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
+++ b/charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
@@ -168,6 +168,7 @@ class TestK8sdAPIManager(unittest.TestCase):
 
         with self.assertRaises(K8sdConnectionError):
             self.api_manager.check_k8sd_ready()
+
         mock_send_request.assert_has_calls(
             [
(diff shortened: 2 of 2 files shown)


charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -776,9 +776,30 @@ def is_cluster_ready(self):
...
     def check_k8sd_ready(self):
-        """Check if k8sd is ready."""
-        endpoint = "/cluster/1.0/ready"
-        self._send_request(endpoint, "GET", EmptyResponse)
+        """Check if k8sd is ready using various microcluster endpoints.
+
+        Raises:
+            K8sdConnectionError: If the response is Not Found on all endpoints.
+        """
+        ready_endpoints = ["/core/1.0/ready", "/cluster/1.0/ready"]
```
addyess, line 784: We'll try the "new one" first -- then back up with the old one

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -776,9 +776,30 @@ def is_cluster_ready(self):
...
+                break
+            except InvalidResponseError as ex:
+                if ex.code == 404:
+                    logger.warning(
+                        "Encountered 404 while checking if micro-cluster is ready @ %s: %s",
+                        endpoint,
+                        ex,
+                    )
+                    # Try the next endpoint if the current one is not found
+                    continue
```
addyess, line 797: k8sd responding with a 404 is its way of saying "I see that request -- but that doesn't exist".  A clear sign we should try another endpoint!

charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
```diff
@@ -143,6 +143,53 @@ def setUp(self):
...
+                call("/cluster/1.0/ready", "GET", EmptyResponse),
+            ]
+        )
+        assert ie.exception.code == 504
+
+    @patch("lib.charms.k8s.v0.k8sd_api_manager.K8sdAPIManager._send_request")
+    def test_check_k8sd_not_found(self, mock_send_request):
+        """Test bootstrap."""
+        not_found = InvalidResponseError(code=404, msg="Not Found")
+        mock_send_request.side_effect = [not_found, not_found]
```
addyess, line 167: A wise man once said...
> two wrongs don't make a right, but three lefts do

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
eaudetcobello: So... this may be misleading the user into thinking something is wrong with the code.
Could the log be modified to mention this *could* be normal and we're cycling through possible endpoints?

charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
```diff
@@ -143,6 +143,53 @@ def setUp(self):
...
+
+        with self.assertRaises(K8sdConnectionError):
+            self.api_manager.check_k8sd_ready()
+        mock_send_request.assert_has_calls(
```
eaudetcobello, line 171:
```suggestion

        mock_send_request.assert_has_calls(
```

charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
```diff
@@ -143,6 +143,53 @@ def setUp(self):
...
+    @patch("lib.charms.k8s.v0.k8sd_api_manager.K8sdAPIManager._send_request")
+    def test_check_k8sd_in_error(self, mock_send_request):
+        """Test bootstrap."""
+        not_found = InvalidResponseError(code=404, msg="Not Found")
+        in_error = InvalidResponseError(code=504, msg="In Error")
+        mock_send_request.side_effect = [not_found, in_error]
+
+        with self.assertRaises(InvalidResponseError) as ie:
+            self.api_manager.check_k8sd_ready()
+        mock_send_request.assert_has_calls(
```
eaudetcobello, line 155:
```suggestion
        
        mock_send_request.assert_has_calls(
```

charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
```diff
@@ -143,6 +143,53 @@ def setUp(self):
...
+    def test_check_k8sd_ready(self, mock_send_request):
+        """Test bootstrap."""
+        not_found = InvalidResponseError(code=404, msg="Not Found")
+        success = EmptyResponse(status_code=200, type="test", error_code=0)
+        mock_send_request.side_effect = [not_found, success]
+
+        self.api_manager.check_k8sd_ready()
+        mock_send_request.assert_has_calls(
+            [
+                call("/core/1.0/ready", "GET", EmptyResponse),
```
eaudetcobello, line 188: Shouldn't this be InvalidResponseError? Because of `mock_send_request.side_effect = [not_found, success]` I would expect the first one to be not found, but I may be misunderstanding.

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
eaudetcobello: Maybe `Trying endpoint 1 of 2: ...`

charms/worker/k8s/tests/unit/test_k8sd_api_manager.py
addyess: In this test case the first endpoint isn't found, but the second is.  Because the second succeeds -- there's no exception right?

eaudetcobello: EmptyResponse is defined as having a 200 status code. If the first endpoint isn't found, why are we expecting a 200 status code?

addyess: Ahh... no 

`EmptyResponse` is the expected "happy" return type.  We'd like to send a request, and get an `EmptyResponse` back. That "third" argument is the expected return type.  

So, yeah `self._send_request(...)` wants to see if the cluster is ready, and see an `EmptyResponse` instance come back.

If we get a status code in the 2xx range, we'll parse the response, otherwise we get an `InvalidResponseError`

eaudetcobello (approved):

PR #149, opened 2024-09-20
Support a charm resource to override the installed snap

addyess:
### Overview

Create support to override the snap, while warning users they are using the charm in an override mode

### Rationale

In order to test that new builds of the k8s snap will operate on the current stable k8s charm, we should be able to verify the snap before releasing it.  To do so, we need a way to insert the snap into the charmed environment.  This provides the ability to do so.


addyess:
We can't merge these changes until the workflows support file-resources on publish

charms/worker/k8s/src/snap.py
```diff
@@ -91,25 +94,140 @@ def _validate_revision(cls, value: Union[str, int, None]) -> Optional[str]:
...
+    log.info("Resource path size: %d bytes", resource_size)
+    unpack_path = _overridden_snap_installation().parent
+    shutil.rmtree(unpack_path, ignore_errors=True)
+    if resource_size == 0:
+        log.info("Resource size is zero bytes. Use the charm defined snap installation script")
+        return _default_snap_installation()
+
+    # Unpack the snap-installation resource
+    unpack_path.mkdir(parents=True, exist_ok=True)
+    command = f"tar -xzvf {resource_path} -C {unpack_path} --no-same-owner"
```
mateoflorido, line 173: Is there any reason to avoid Python's tar [library](https://docs.python.org/es/3/library/tarfile.html)?

charms/worker/k8s/src/snap.py
```diff
@@ -91,25 +94,140 @@ def _validate_revision(cls, value: Union[str, int, None]) -> Optional[str]:
...
+                {
+                    "install-type": "file",
+                    "name": "k8s",
+                    "filename": str(snap_path[0]),
+                    "classic": True,
+                    "dangerous": True,
+                }
+            ]
+        }
+        yaml.safe_dump(manifest, snap_installation.open("w"))
```
mateoflorido, line 202:
```suggestion
        with snap_installation.open("w") as f:
            yaml.safe_dump(manifest, f)
```
Wdyt?

charms/worker/k8s/src/snap.py
```diff
@@ -126,24 +244,63 @@ def _parse_management_arguments() -> List[SnapArgument]:
...
+                which.ensure(**install_args)
                 which.hold()
         elif isinstance(args, SnapStoreArgument):
             log.info("Ensuring %s snap channel=%s", args.name, args.channel)
-            which.ensure(**args.dict(exclude_none=True))
+            which.ensure(**install_args)
 
 
-def version(snap: str) -> Optional[str]:
+def block_refresh(which: snap_lib.Snap, args: SnapArgument) -> bool:
```
mateoflorido, line 271: I was wondering, is there a way to allow the installation? For example, if we're upgrading, can we instruct the snap module to proceed with the upgrade?

charms/worker/k8s/src/charm.py
```diff
@@ -110,6 +110,38 @@ class NodeRemovedError(Exception):
...
+
+    @message.setter
+    def message(self, message: str):
+        """Set the message for the status.
+
+        Args:
+            message (str): explanation of the unit status
+        """
+        self._message = message
+
```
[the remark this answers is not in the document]
addyess: Done, look at `events.update_status`

charms/worker/k8s/src/snap.py
addyess: it wasn't easy, but is done

charms/worker/k8s/src/snap.py
addyess: Also wasn't easy but done!

mateoflorido (approved):
LGTM! :shipit:

PR #154, opened 2024-10-20
Add Annotations to Charm Configuration

### LINKED ISSUE #112: Ability to configure cilium option --vlan-bpf-bypass
### Enhancement Proposal

Cilium failed to start when the primary bond has more than 5 VLANs.
This requires to configure cilium charm --vlan-bpf-bypass option.

[1] https://docs.cilium.io/en/stable/configuration/vlan-802.1q/

mateoflorido:
## Overview
Add a charm configuration option to enable the fine grained configuration of clusters using annotations.
## Rationale
The `k8s` snap allows the use of annotations as a mechanism to configure features built into the snap. For example, to configure VLAN BPF bypass in Cilium (as requested in #112), we can use the annotation `k8sd/v1alpha1/cilium/vlan-bpf-bypass` which passes the VLAN IDs to Cilium's configuration.
## Related Issues
closes #112

### DIFF
--- a/charms/worker/k8s/charmcraft.yaml
+++ b/charms/worker/k8s/charmcraft.yaml
@@ -56,9 +56,9 @@ config:
   options:
     annotations:
       type: string
-      default: "{}"
+      default: ""
       description: |
-        Annotations is a JSON dictionary (key/value pairs) object that can be
+        Annotations is a space separated list of (key/value) pairs)  that can be
         used to add arbitrary metadata configuration to the Canonical
         Kubernetes cluster. For more information, see the upstream Canonical
         Kubernetes documentation about annotations:
@@ -66,7 +66,7 @@ config:
         https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/reference/annotations/
 
         The value for this config must be a JSON object, like this:
-          e.g.: {"key1": "value1", "key2": "value2"}
+          e.g.: key1=value1 key2=value2
     containerd_custom_registries:
       type: string
       default: "[]"
--- a/charms/worker/k8s/charmcraft.yaml
+++ b/charms/worker/k8s/charmcraft.yaml
@@ -65,7 +65,7 @@ config:
 
         https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/reference/annotations/
 
-        The value for this config must be a JSON object, like this:
+        Example:
           e.g.: key1=value1 key2=value2
     containerd_custom_registries:
       type: string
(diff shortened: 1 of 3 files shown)


bschimke95 (approved):
LGTM! Great work @mateoflorido

charms/worker/k8s/charmcraft.yaml
```diff
@@ -54,6 +54,19 @@ bases:
...
+      description: |
+        Annotations is a JSON dictionary (key/value pairs) object that can be
+        used to add arbitrary metadata configuration to the Canonical
+        Kubernetes cluster. For more information, see the upstream Canonical
+        Kubernetes documentation about annotations:
+
+        https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/reference/annotations/
+
+        The value for this config must be a JSON object, like this:
+          e.g.: {"key1": "value1", "key2": "value2"}
```
addyess, line 69: are the annotations ever `key=value` pairs where key and value are non-strings?  
do the values for `key` or `value` ever contain the `=` character?  

If the answer to both of these is no, may i recommend:

```suggestion
      default: ""
      description: |
        Annotations is a space separated list of (key/value) pairs)  that can be
        used to add arbitrary metadata configuration to the Canonical
        Kubernetes cluster. For more information, see the upstream Canonical
        Kubernetes documentation about annotations:

        https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/reference/annotations/

        The value for this config must be a JSON object, like this:
          e.g.: key1=value1 key2=value2
```

addyess: Also, can you say what resources we're specifically annotating?  Are they only the worker nodes?  This PR only has changed the k8s-worker charms' charmcraft.yaml?

addyess (changes requested):
I think the impl is right -- just have curiosity about what is being annotated and if the control-plane needs it too

charms/worker/k8s/charmcraft.yaml
addyess: OH OH OH... these are microcluster annotations.  I understand now that these are shared across the whole cluster using microcluster

charms/worker/k8s/src/charm.py
```diff
@@ -329,6 +330,66 @@ def _configure_cos_integration(self):
...
+        annotations = self._get_valid_annotations()
+        if annotations is None:
+            return
+
+        if not config.cluster_config:
+            config.cluster_config = UserFacingClusterConfig(annotations=annotations)
+        else:
+            config.cluster_config.annotations = annotations
+
+    def _update_annotations(self):
```
addyess, line 377: does this need a catch for the failure condition?

```suggestion
    @status.error(...)
    def _update_annotations(self):
```

addyess (approved):
🚀 Lets go!

PR #160, opened 2024-11-06
Subordinate charm config PoC

eaudetcobello:
Applicable spec: [Internal spec](https://docs.google.com/document/d/1FeGPsiOTjAs2_QBHJcrVtmw4YePL4-d-Swe4m5kIlq8/)

Goes with: https://github.com/eaudetcobello/k8s-load-balancer-poc/blob/main/src/charm.py

### Overview

Adds the `feature` interface for subordinate charms to attach to. Subordinate charms can change relation data to communicate with k8s-operator and configure cluster features.

### Rationale

The `feature` interface and observing the relation-changed event is needed to act when a subordinate charm is related to k8s-operator.

### Juju Events Changes

Observe feature-relation-changed events.

### Module Changes

None.

### Library Changes

None.

### Checklist

- [ ] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [ ] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)


charms/worker/k8s/src/charm.py
```diff
@@ -149,10 +151,47 @@ def __init__(self, *args):
         )
 
         self.framework.observe(self.on.update_status, self._on_update_status)
+        self.framework.observe(self.on.feature_relation_changed, self._on_feature_relation_changed)
+
```
addyess, line 155: so, with the reconciler pattern in this charm, registering for each event isn't necessary.  You're going to see every event the charm can possible observe in the `self._reconcile` method.  

It's likely best to have this `_on_feature_relation_changed` method be named `_resolve_feature_relations` and called from `self._reconcile`

charms/worker/k8s/src/charm.py
```diff
@@ -149,10 +151,47 @@ def __init__(self, *args):
...
         if self.is_control_plane:
             self.etcd = EtcdReactiveRequires(self)
             self.framework.observe(self.on.get_kubeconfig_action, self._get_external_kubeconfig)
 
+    def _on_feature_relation_changed(self, event: ops.RelationChangedEvent):
```
addyess, line 160: so, once this is called from the `reconciler(...)` -- the signature will change a bit.  We won't know which event we're responding to -- it may not even be a relation event.  What we should check instead is 
* are there any feature relations
* do those feature relations have unit data
* can any of the relation data be interpreted as snap config

charms/worker/k8s/src/charm.py
addyess: oh, also this should only be run on the "leader-control-plane" unit.  no where else

charms/worker/k8s/src/charm.py
```diff
@@ -149,10 +151,47 @@ def __init__(self, *args):
...
+        relation_data = event.relation.data.get(event.unit) #type: ignore
+        if not relation_data:
+            log.warning("No relation data found for unit %s", event.unit)
+            return
+
```
addyess, line 165: likely this will loop over all the feature relations here.

```python
for relation in self.model.relations["feature"]:
    for unit in relation.units:
        relation_data = relation.data[unit]
```

redeploy.sh
```diff
@@ -0,0 +1,28 @@
+set -x
```
addyess, line 1: This might be a handy script for deploying this charm over and over -- but let's exclude from the PR for the time being

charms/worker/k8s/src/charm.py
```diff
@@ -149,10 +151,47 @@ def __init__(self, *args):
...
+            "load-balancer": LoadBalancerConfig,
+            "local-storage": LocalStorageConfig,
+        }
+
+        if feature_name and feature_version and feature_attributes:
+            feature_attributes = json.loads(feature_attributes)
+
+            config_class = feature_config_classes.get(feature_name)
+            if config_class:
+                feature_config = config_class(**feature_attributes)
```
addyess, line 180: I feel like pydantic should know how to unmarshall a json string into one of it's classes.

charms/worker/k8s/src/charm.py
```diff
@@ -149,10 +151,47 @@ def __init__(self, *args):
...
+
+                feature_name = feature_name.replace("-", "_")
+                config = UserFacingClusterConfig(**{feature_name: feature_config})
+                update_request = UpdateClusterConfigRequest(config=config)
+                self.api_manager.update_cluster_config(update_request)
+
+                log.info("Got feature of type [%s], with version [%s] and attributes [%s]", feature_name, feature_version, json.dumps(feature_config.dict()))
+            else:
+                log.warning("Unsupported feature name: %s", feature_name)
```
addyess, line 189: Here's a good situation to "block the charm" 
you can do this with `status.add(ops.BlockedStatus(f"Unsupported feature {feature_name}"))`

addyess (changes requested):
Alrighty... i provided you some places to improve.  But yeah this is coming along

mateoflorido (changes requested):
Great work! I have a few suggestions

charms/worker/k8s/src/charm.py
```diff
@@ -149,10 +151,47 @@ def __init__(self, *args):
...
 
+    def _on_feature_relation_changed(self, event: ops.RelationChangedEvent):
+        relation_data = event.relation.data.get(event.unit) #type: ignore
+        if not relation_data:
+            log.warning("No relation data found for unit %s", event.unit)
+            return
+
+        feature_name = relation_data.get("feature-name")
+        feature_version = relation_data.get("feature-version") # library version
+        feature_attributes = relation_data.get("feature-attributes")
```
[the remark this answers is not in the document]
eaudetcobello: Done, lmk what you think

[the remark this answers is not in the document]
eaudetcobello: Thanks a lot for the idea. I implemented a basic version of it (checking some fields, but not version) just because this is a PoC and I need to move on to other things!

Issue #110, opened 2024-05-13
Enable and configure LocalStorage feature


hemanthnakkina:
### Enhancement Proposal

Wishlist to support enabling and configuring localstorage feature either via k8s-operator or separate charm

ktsakalozos:
@hemanthnakkina do you want to configure local storage of do you want it enabled by default?

hemanthnakkina:
Currently enabling by default should be sufficient as sunbeam does not update any configuration provided by local storage.

mateoflorido:
Fixed: #169

Issue #109, opened 2024-05-13
Enable and configure loadbalancer feature


hemanthnakkina:
### Enhancement Proposal

Wishlist to support enabling and configuring  loadbalancer feature via k8s-operator charm or separate charm.

PR #169, opened 2024-11-18
Add Local storage configuration through charm config

bschimke95:
### Overview

Note: The naming might change and is WIP for now.
Add options to the `k8s` charm to configure the local storage via charm configuration.

### Rationale

A user may want to overwrite the default configuration for local storage. This should be possible both - when bootstrapping a cluster and during runtime.

### Module Changes

* Refactor the existing annotation reconfiguration to support the whole `UserfacingClusterConfig`
* Implement the local-storage part of the charm configuration - the remaining configs are updated in follow-up work.


charms/worker/k8s/src/charm.py
```diff
@@ -372,41 +372,59 @@ def _get_valid_annotations(self) -> Optional[dict]:
...
-            config.cluster_config = UserFacingClusterConfig(annotations=annotations)
+            config.cluster_config = self._get_cluster_config_from_charm_config()
         else:
-            config.cluster_config.annotations = annotations
+            charm_cluster_config = self._get_cluster_config_from_charm_config()
+            # Merge the charm and user cluster configurations
+            # The existing configuration takes precedence over the charm configuration as
+            # this information is retrieved elsewhere (e.g. cloud-provider).
+            merged_data = {**charm_cluster_config.dict(), **config.cluster_config.dict()}
+            config.cluster_config = UserFacingClusterConfig(**merged_data)
```
bschimke95, line 407: I don't know if I'm super happy with this kind of merging of values as it somewhat is non-intuitive that the existing config takes precedence. I adapted here to the existing code.

charms/worker/k8s/charmcraft.yaml
```diff
@@ -161,6 +161,32 @@ config:
...
+      description: |
+        The reclaim policy for local storage. This can be either "Delete" or
+        "Retain". If set to "Delete", the storage will be deleted when the
+        PersistentVolumeClaim is deleted. If set to "Retain", the storage will
+        be retained when the PersistentVolumeClaim is deleted.
+    local_storage_set_default:
+      type: boolean
+      description: |
+        Set the local storage class as the default storage class. This is only
+        used if local-storage.enabled is set to true.
```
eaudetcobello, line 189: Should be local-storage-enabled to match existing config options.

addyess, line 189: Came to same something similar here:

1) as discussed lets stick to `snake_case` config names.  That's good
2) let's namespace this with the feature name:  `storage_local_*`

charms/worker/k8s/charmcraft.yaml
```diff
@@ -161,6 +161,32 @@ config:
         CIDR to use for Kubernetes services. After deployment it is
         only possible to increase the size of the IP range. It is not possible to
         change or shrink the address range after deployment.
+    local_storage_enabled:
+      type: boolean
```
addyess, line 165: let's establish a default value.  I *THINK* the request was it's enabled by default.  so please add a `default: true`

charms/worker/k8s/charmcraft.yaml
```diff
@@ -161,6 +161,32 @@ config:
...
+      description: |
+        Enable local storage provisioning. This will create a storage class
+        named "local-storage" that uses the hostPath provisioner. This is
+        useful for development and testing purposes. It is not recommended for
+        production use.
+    local_storage_local_path:
+      type: string
+      default: ""
```
addyess, line 173: let's go ahead and copy through the default value from the snap shall we?  is that ""?  What path does this end up consuming?

charms/worker/k8s/charmcraft.yaml
```diff
@@ -161,6 +161,32 @@ config:
...
+        path must be writable by the kubelet. This is only used if
+        local-storage.enabled is set to true.
+    local_storage_reclaim_policy:
+      type: string
+      description: |
+        The reclaim policy for local storage. This can be either "Delete" or
+        "Retain". If set to "Delete", the storage will be deleted when the
+        PersistentVolumeClaim is deleted. If set to "Retain", the storage will
+        be retained when the PersistentVolumeClaim is deleted.
+    local_storage_set_default:
```
addyess, line 185: i don't like this config because it's meaningless. 

what if i set it to true?  Ok, so at that moment, it will set the local storage to the default.  If a kube admin comes along and changes it to a new storage class with `kubectl`-- well then THAT SC be comes the new default.  But this config isn't changed -- its still `true` isn't it?  Now we've got a UX problem.  

Can this config item just be dropped?

charms/worker/k8s/charmcraft.yaml
```diff
@@ -161,6 +161,32 @@ config:
...
+        named "local-storage" that uses the hostPath provisioner. This is
+        useful for development and testing purposes. It is not recommended for
+        production use.
+    local_storage_local_path:
+      type: string
+      default: ""
+      description: |
+        The path on the host where local storage will be provisioned. This
+        path must be writable by the kubelet. This is only used if
+        local-storage.enabled is set to true.
```
addyess, line 177:
```suggestion
        storage_local_enabled is set to true.
```

charms/worker/k8s/src/charm.py
```diff
@@ -133,7 +133,7 @@ def __init__(self, *args):
         self.labeller = LabelMaker(
             self, kubeconfig_path=self._internal_kubeconfig, kubectl=KUBECTL_PATH
         )
-        self._stored.set_default(is_dying=False, cluster_name=str(), annotations={})
```
addyess, line 136: why is this charm stored state?  This doesn't feel right.  I'm not sure why the annotations are stored state either.  Can't we just fetch this from the k8s snap?

charms/worker/k8s/src/charm.py
```diff
@@ -372,41 +372,59 @@ def _get_valid_annotations(self) -> Optional[dict]:
...
+        status.add(ops.MaintenanceStatus("Updating cluster configuration"))
+        log.info("Updating cluster configuration")
 
-        status.add(ops.MaintenanceStatus("Updating Annotations"))
-        log.info("Updating Annotations")
+        cluster_config = self._get_cluster_config_from_charm_config()
 
-        if self._stored.annotations == annotations:
+        if self._stored.cluster_config == cluster_config:
             return
```
addyess, line 421: Yeah, this isn't very good.  I don't like that we're storing snap config in the charm.  It can get off sync. 

we should fetch the cluster config from the k8s snap, and compare it with the new charm config.  If they differ we need to update. 

Storing this info in the charm's stored state isn't the right path.

addyess (changes requested):
Hopefully the comments provide details needed to complete. Let's talk if they dont

charms/worker/k8s/charmcraft.yaml
ktsakalozos: We should be using dashes and not underscores, see https://juju.is/docs/sdk/styleguide#heading--naming

bschimke95: naming is hard...

I now use the dashes and namespace it as `local_storage`
We still need to decide on a prefix as discussed in MM.

charms/worker/k8s/charmcraft.yaml
bschimke95: Done

charms/worker/k8s/charmcraft.yaml
bschimke95: dropped it

charms/worker/k8s/charmcraft.yaml
bschimke95: As written in MM, I think `storage_local` is kinda weird. Is there a reason you want to  switch the order?

charms/worker/k8s/src/charm.py
bschimke95: I adapted here to the existing annotation code but agree - it is weird. Removed it and directly query the API now.

charms/worker/k8s/src/charm.py
bschimke95: updated

charms/worker/k8s/charmcraft.yaml
bschimke95: as agreed on in the stand-up: `local-storage` it is

charms/worker/k8s/charmcraft.yaml
bschimke95: no prefix, as discussed in the stand-up

charms/worker/k8s/src/charm.py
addyess: Tell me more here.  the `config.cluster_config` is what the snap currently understands
`charm_cluster_config` is the juju admin's view of the configuration. 

why does the snap config take precedence over the juju admin config?

addyess (approved):
LGTM Just have a question now

charms/worker/k8s/src/charm.py
```diff
@@ -383,39 +383,62 @@ def _get_valid_annotations(self) -> Optional[dict]:
...
     )
-    def _update_annotations(self):
-        """Update the annotations for the Canonical Kubernetes cluster."""
-        annotations = self._get_valid_annotations()
-        if annotations is None:
-            return
+    def _update_cluster_config(self):
+        """Update the cluster configuration for the Canonical Kubernetes cluster."""
+        status.add(ops.MaintenanceStatus("Updating cluster configuration"))
+        log.info("Updating cluster configuration")
```
mateoflorido, line 428: Can we move this after we know we are updating the cluster config? I'd like to see less flapping in the juju status message.

charms/worker/k8s/src/charm.py
```diff
@@ -383,39 +383,62 @@ def _get_valid_annotations(self) -> Optional[dict]:
...
 
-    def _configure_annotations(self, config: BootstrapConfig):
-        """Configure the annotations for the Canonical Kubernetes cluster.
+    def _get_cluster_config_from_charm_config(self) -> UserFacingClusterConfig:
+        """Retrieve the cluster config from charm configuration.
+
+        Returns:
+            UserFacingClusterConfig: The expected cluster configuration.
+        """
+        local_storage = LocalStorageConfig(
```
HomayoonAlimohammadi, line 392: I think we need to add `allow_population_by_field_name = True` to the `LocalStorageConfigModel`. Otherwise _I think_ the fields with aliases (e.g. `local_path`) are not going to get populated like this: https://programiz.pro/ide/python/ORE3ZZSJRG?utm_medium=playground&utm_source=python_playground-shared-project-link
```python
class LocalStorageConfig(BaseModel):
    """Configuration for the local storage settings of the cluster.

    Attributes:
        enabled: Optional flag which represents the status of Storage.
        local_path: The local path for storage.
        reclaim_policy: The policy for reclaiming local storage.
        set_default: Optional flag to set this as the default storage option.
    """

    enabled: Optional[bool] = Field(None)
    local_path: Optional[str] = Field(None, alias="local-path")
    reclaim_policy: Optional[str] = Field(None, alias="reclaim-policy")
    set_default: Optional[bool] = Field(None, alias="set-default")

    class Config:
        """Pydantic configuration for the LocalStorageConfig model."""
        allow_population_by_field_name = True
```

p.s. Wish we had an easy Python playground just like Go...

HomayoonAlimohammadi: [GIVEN THAT I'M CORRECT] I found this accidentally trying to understand why is Pylance going crazy and underlining everything in red. 
Maybe we need to add a test that populates and validates this config so that we don't miss anything.

charms/worker/k8s/src/charm.py
```diff
@@ -383,39 +383,62 @@ def _get_valid_annotations(self) -> Optional[dict]:
...
-            return
+    def _update_cluster_config(self):
+        """Update the cluster configuration for the Canonical Kubernetes cluster."""
+        status.add(ops.MaintenanceStatus("Updating cluster configuration"))
+        log.info("Updating cluster configuration")
 
-        status.add(ops.MaintenanceStatus("Updating Annotations"))
-        log.info("Updating Annotations")
+        charm_cluster_config = self._get_cluster_config_from_charm_config()
+        snap_cluster_config = self.api_manager.get_cluster_status().metadata.status.config
```
HomayoonAlimohammadi, line 431: Seems like `.metadata` can be `None`. Maybe we can check that to prevent exceptions?

bschimke95: I think that should be fine and is as designed.
The reconciler loop will just fail and retry, am I correct @mateoflorido ?

bschimke95:
Works as expected:
image

charms/worker/k8s/src/charm.py
bschimke95: Thanks - for some reason it worked anyway (maybe the settings from `UserfacingConfig` are applied recursively?

Anyway, added it to be explicit.

charms/worker/k8s/src/charm.py
bschimke95: Removed this function in favour of the existing `ensure_cluster_config`
but you were totally right.

charms/worker/k8s/src/charm.py
bschimke95: That being said, it is outdated now anyway - we can safely send the update request to the snap and the snap will only perform any actions if something actually changed.

charms/worker/k8s/src/charm.py
bschimke95: discussed offline - the code structure was not great and I updated it now to be more concise.

charms/worker/k8s/src/charm.py
```diff
@@ -792,7 +759,6 @@ def _reconcile(self, event: ops.EventBase):
             self._k8s_info(event)
             self._bootstrap_k8s_snap()
             self._enable_functionalities()
```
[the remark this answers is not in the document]
HomayoonAlimohammadi: Resolved for now. Can be addressed later after all the features have been merged.
Created a card for it.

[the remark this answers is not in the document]
bschimke95: good point but this can only be removed after all related PRs are merged.

addyess (approved):
LGTM too

mateoflorido:
closes #109

mateoflorido:
closes #110

PR #170, opened 2024-11-18
 Add Load Balancer configuration through charm config

eaudetcobello:
### Overview

Note: The naming might change and is WIP for now.
Add options to the k8s charm to configure the load balancer via charm configuration.

### Rationale

A user may want to overwrite the default configuration for load balancer. This should be possible both - when bootstrapping a cluster and during runtime.

### Module Changes

* Implement the load-balancer part of the charm configuration - the remaining configs are updated in follow-up work.


bschimke95:
This is missing the actual mapping of the options in `charm.py`.
I'd say let's wait for the local-storage PR to land, then this is a quick change.

charms/worker/k8s/charmcraft.yaml
```diff
@@ -189,6 +189,54 @@ config:
...
+      description: |
+        Enable/Disable L2 mode for the load balancer. This is only used if
+        load-balancer-enabled is set to true.
+    load-balancer-l2-interfaces:
+      type: string
+      default: ""
+      description: |
+        Space-separated list of interfaces to use for the load balancer. This
+        is only used if load-balancer-l2-mode is set to true. if unset, all
+        interfaces will be used.
```
addyess, line 215: Something feels odd here. I suppose it's likely that all the interfaces used for the L2 load balancer will be the same on all the machines. But this could be better if it was a regex rather than a fixed list of interfaces not really known until the machine is deployed

addyess (commented):
A few questions for @bschimke95 and @eaudetcobello

charms/worker/k8s/charmcraft.yaml
eaudetcobello: This is valid, it could be a feature we add in the future. I have no idea if MetalLB supports regex for the interfaces.

See: https://metallb.universe.tf/configuration/_advanced_l2_configuration/index.html#specify-network-interfaces-that-lb-ip-can-be-announced-from

addyess: So, we could do the regex IN the charm.  It doesn't necessarily have to apply to metallb.  We've got to remain generic here

addyess: Lets leave this regex idea as a future feature

addyess (approved):
LGTM

mateoflorido:
closes #109

PR #173, opened 2024-11-19
Replace AssertionError with ReconcileError

HomayoonAlimohammadi:
### Overview

Replace `AssertionError` with a newly defined `ReconcileError`.

### DIFF
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -191,7 +191,6 @@ class K8sCharm(ops.CharmBase):
     @status.on_error(
         ops.WaitingStatus("Installing COS requirements"),
         subprocess.CalledProcessError,
-        AssertionError,
     )
     def _apply_cos_requirements(self):
         """Apply COS requirements for integration.
@@ -315,7 +314,7 @@ class K8sCharm(ops.CharmBase):
 
     @on_error(
         ops.WaitingStatus("Waiting to bootstrap k8s snap"),
-        AssertionError,
+        ReconcilerError,
         InvalidResponseError,
         K8sdConnectionError,
     )
@@ -387,7 +386,7 @@ class K8sCharm(ops.CharmBase):
             dict: The parsed annotations if valid, otherwise None.
 
         Raises:
-            AssertionError: If any annotation is invalid.
+            ReconcilerError: If any annotation is invalid.
         """
         raw_annotations = self.config.get("annotations")
         if not raw_annotations:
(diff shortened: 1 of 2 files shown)


charms/worker/k8s/src/errors.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Custom exceptions for the worker charm."""
+
+class ReconcileError(Exception):
+    """Exception raised for errors that occur during the reconciliation process."""
+    ...
```
addyess, line 8: hold up... this exception is already defined...

you'll find it here 
```from charms.contextual_status import ReconcilerError```


I think the ask here is to not use this style in the code:

 ```python
assert <truthy statement>, "Message about the truthy statement"
```

but rather:

```python
if not <truthy_statement>:
    raise ReconcilerError("Message about the truthy statement")
```

addyess (changes requested):
Hey @HomayoonAlimohammadi thanks for taking a stab at this.  Let me help clarify the ask on this task

charms/worker/k8s/src/errors.py
HomayoonAlimohammadi: That's makes a lot more sense...
Done. Thanks!

charms/worker/k8s/src/charm.py
```diff
@@ -184,7 +184,7 @@ def _k8s_info(self, event: ops.EventBase):
     @status.on_error(
         ops.WaitingStatus("Installing COS requirements"),
         subprocess.CalledProcessError,
-        AssertionError,
+        ReconcilerError,
```
mateoflorido, line 187: Hmmm, this method doesn't raise a ReconcilerError, so it should be fine to remove it.

mateoflorido (approved):
LGTM! Just a nitpick

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: Thanks! Done!

charms/worker/k8s/src/token_distributor.py
```diff
@@ -246,14 +257,17 @@ def recover_token(self, relation: ops.Relation) -> Generator[str, None, None]:
...
-        assert secret_id, f"{relation.name}:{secret_key} is not valid"  # nosec
+        if not secret_id:
+            raise ReconcilerError(f"{relation.name}:{secret_key} is not valid")
         secret = self.charm.model.get_secret(id=secret_id)
 
         # Get the content from the secret
         content = secret.get_content(refresh=True)
-        assert content["token"], f"{relation.name}:token not valid"  # nosec
+        if content.get("token") is None:
+            raise ReconcilerError(f"{relation.name}:token not valid")
```
addyess, line 270: i think this is more accurate

```suggestion
        if not content.get("token"):
            raise ReconcilerError(f"{relation.name}:token not valid")
```

addyess (approved):
Please fix tiny nit... then merge with passing tests

charms/worker/k8s/src/token_distributor.py
HomayoonAlimohammadi: Thanks! Fixed this. Just as a question: Are we changing this to also include content.get("token") == ""?

PR #175, opened 2024-11-19
Add Network feature config options

HomayoonAlimohammadi:
### Overview

Add Network feature config options

### Rationale

Users should be able to configure network-related options on the charm.

### Note for reviewers

Should be rebased and merged after https://github.com/canonical/k8s-operator/pull/169

The description of PodCIDR and ServiceCIDR contains this:
`Can contain at most 2 blocks, one for IPv4 and one for IPv6.`
This is not "truly" accurate according to how we parse the CIDR string: https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/utils/cidr.go#L169-L192
According to the snippet above, we're technically allowed to pass 2 blocks of the same type (v4/v6) but _I suspect_ that's not what we want to have. We should check that if there are two CIDRs, one is v4 and the other one is v6.


charms/worker/k8s/charmcraft.yaml
```diff
@@ -161,6 +161,62 @@ config:
...
+        path must be writable by the kubelet. This is only used if
+        local-storage.enabled is set to true.
+    local-storage-reclaim-policy:
+      type: string
+      default: Delete
+      description: |
+        The reclaim policy for local storage. This can be either "Delete" or
+        "Retain". If set to "Delete", the storage will be deleted when the
+        PersistentVolumeClaim is deleted. If set to "Retain", the storage will
+        be retained when the PersistentVolumeClaim is deleted.
```
addyess, line 186: this encroaches on https://github.com/canonical/k8s-operator/pull/169

addyess: Is this in addition to ben's work somehow?

bschimke95: I think Hue just branched out from my PR and just forgot to set the base branch in GH from `main` to `whatever-my-branch-is-called`

HomayoonAlimohammadi: That's right, sorry I had to also change the base.

charms/worker/k8s/charmcraft.yaml
```diff
@@ -154,13 +154,66 @@ config:
...
+        and one for IPv6.
+        After deployment it is only possible to increase the size of 
+        the IP range. It is not possible to change or shrink the address 
+        range after deployment.
+
+        Examples: 
+          - "192.0.2.0/24"
+          - "2001:db8::/32"
+          - "192.0.2.0/24,2001:db8::/32"
+          - "2001:db8::/32,192.0.2.0/24"
```
addyess, line 172: I'm going to add pod-cidr in #179  But i'm going to steal these notes

pyproject.toml
```diff
@@ -86,4 +86,4 @@ max-complexity = 10
 skip = "build,lib,venv,icon.svg,.tox,.git,.mypy_cache,.ruff_cache,.coverage"
 
 [tool.pyright]
-extraPaths = ["./charms/worker/k8s/lib"]
+extraPaths = ["./charms/worker/k8s/lib", "./charms/worker/k8s/src"]
```
HomayoonAlimohammadi, line 89: This is so that `from charm import K8sCharm` gets resolved by Pylance in the test file.

addyess (approved):
Looks perfect.  Ship it!

PR #176, opened 2024-11-19
Add Ingress feature configuration

HomayoonAlimohammadi:
### Overview

Add ingress feature config options

### Rationale

Users should be able to configure ingress feature options on the charm.

### Note to reviewers

Should be rebased and merged after https://github.com/canonical/k8s-operator/pull/175

- Just a reminder that if the `default-tls-secret` is not given it will be an empty string which is being used while applying ingress manifests in `k8s-snap`:
    - https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/features/contour/ingress.go#L126-L146
    - https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/features/cilium/ingress.go#L29-L38


charms/worker/k8s/charmcraft.yaml
```diff
@@ -214,6 +214,25 @@ config:
...
+      type: boolean
+      default: false
+      description: |
+        Determines if the ingress feature should be enabled.
+    ingress-default-tls-secret:
+      type: string
+      default: ""
+      description: |
+        Sets the name of the secret to be used for providing default encryption to
+        ingresses.
```
addyess, line 227: Is this a kubernetes secret?  Can we clarify it's not to be confused with a `juju secret`?  Also, is there a namespace this secret should stay in?  Is it always kube-system or something else?

```suggestion
        Sets the name of the kubernetes secret to be used for providing default encryption to
        ingresses.
```

addyess (commented):
I'd like a bit more description on this for juju administrators to have context

charms/worker/k8s/charmcraft.yaml
HomayoonAlimohammadi: Thanks for pointing this out. I guess the namespace depends on the ingress being used.
For contour (in moonray impl) it's `projectcontour-root`: https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/features/contour/chart.go#L24-L29
For cilium (in default impl) it's `kube-system`: https://github.com/canonical/k8s-snap/blob/main/src/k8s/pkg/k8sd/features/cilium/ingress.go#L30-L38

Since I believe we're going with the default implementation which uses Cilium, it's best to say it is in the `kube-system`? Or maybe we can point them to a reference page about this.

addyess: yucky.  that's sad

if this secret moves between implementations we shouldn't specify which namespace.  We should be clear that's it is closely associated with the ingress implementation.  

Who sets this secret by the way?  Is the snap creating the secret (via the helm charm)?  Why do I want to set this secret name? I'm sorry i have so many questions. 

let's  make sure the description here is generic enough to handle multiple implementations under the hood, but specific enough to let the juju admin know they need to administer something in the cluster. 

if it's useless -- can we just drop the secret?

@bschimke95 do you have any feedback here?

addyess (approved):
LGTM

PR #190, opened 2024-11-26
[Backport 1.31] Fix dns config name (#189)

bschimke95:
Backport of dns config bug fix: https://github.com/canonical/k8s-operator/pull/189


addyess (approved):
Fascinating. Without an integration test we would never find this. The least test would be to change the config and see the config status change under the hood

bschimke95:
I added a card for us to create such a test in the next iteration.

PR #200, opened 2024-12-02
Implement `upgrade-relation` for control plane nodes

mateoflorido:
## Overview

Introduce upgrade orchestration for control plane nodes.

## Rationale
To simplify the upgrade process for charms, this pull request adds orchestration logic specifically for upgrading the control plane nodes, specifically the charm core (`k8s` snap). It does not cover the worker node upgrade orchestration, which will be addressed in a future pull request.

### Changes
- Added the `on_upgrade_granted` handler to manage the upgrade process for nodes in the cluster.
- Integrated the upgrade instance into the status update handler to reflect the current upgrade status in the charm's status.
- Enhanced the snap and charm modules to support unblocking snap installations when the upgrade logic confirms all conditions are met.


charms/worker/k8s/src/charm.py
```diff
@@ -694,8 +722,9 @@ def _announce_kubernetes_version(self):
                 if not unit_version:
                     raise ReconcilerError(f"Waiting for version from {unit.name}")
                 if unit_version != local_version:
-                    status.add(ops.BlockedStatus(f"Version mismatch with {unit.name}"))
-                    raise ReconcilerError(f"Version mismatch with {unit.name}")
+                    # NOTE: Add a check to validate if we are doing an upgrade
+                    status.add(ops.WaitingStatus("Upgrading the cluster"))
+                    return
```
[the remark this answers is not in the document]
addyess: Ahh, i see now that it's because the raising of the reconciler error prevented the upgrade events from engaging.   Whew -- we still need a check to let folks know they're running on out-of-spec-version of the applications

For now `_announce_kubernetes_version` is only run on the lead CP.   say you deployed and related a kw 1.35 to a 1.31 cluster.  I imagine the 1.35 workers may not join.  Should they join?  Is the k8s-cp the right place to gripe about it?  You're right that we should at least make sure we're not in an upgrade scenario before we raise the reconciler error.

addyess (changes requested):
Upgrades are obviously hard.  We want to guide the user rather than road-blocking them at every turn.  In the end, the juju admin knows best what they want their cluster to do and how to get it to the revision they want.  We shouldn't make unlimited hurdles from them to cross.  This process should guide them -- not block them at every turn. 

I say this as one who has performed these upgrades before and putting too many guardrails up saves the casual user but can impede the person who has accepted the risk and just wants the cluster upgraded

addyess (approved):
Awesome work here.  Thanks mateo!

charms/worker/k8s/src/literals.py
```diff
@@ -3,6 +3,20 @@
...
+    "kubelet",
+    "kube-proxy",
+    "k8sd",
+]
+
+K8S_CONTROL_PLANE_SERVICES = [
+    "kube-apiserver",
+    "kube-controller-manager",
+    "kube-scheduler",
+]
```
eaudetcobello, line 18: k8s-dqlite, k8s-apiserver-proxy, containerd seem to be missing.

charms/worker/k8s/src/snap.py
```diff
@@ -272,7 +273,7 @@ def management(charm: ops.CharmBase) -> None:
     cache = snap_lib.SnapCache()
     for args in _parse_management_arguments(charm):
         which = cache[args.name]
-        if block_refresh(which, args):
+        if block_refresh(which, args) and not charm.is_upgrade_granted:
```
eaudetcobello, line 276: Shouldn't this be `or`?

charms/worker/k8s/src/upgrade.py
```diff
@@ -71,13 +106,109 @@ def pre_upgrade_check(self) -> None:
...
-        if failing_pods := self.node_manager.verify_pods_running(["kube-system"]):
+        if failing_pods:
             raise ClusterNotReadyError(
                 message="Cluster is not ready",
                 cause=f"Pods not running in namespace(s): {failing_pods}",
                 resolution="Check the logs for the failing pods.",
             )
 
+    def _verify_worker_version(self) -> bool:
+        """Verify the worker version.
```
eaudetcobello, line 117: Comment does not add much context.

charms/worker/k8s/src/upgrade.py
```diff
@@ -71,13 +106,109 @@ def pre_upgrade_check(self) -> None:
...
+            self.set_unit_failed(cause="Failed to get the version of the k8s snap.")
+            status.add(ops.BlockedStatus("Failed to get the version of the k8s snap."))
+            return
+
+        status.add(ops.MaintenanceStatus("Upgrading the charm."))
+
+        if self.charm.lead_control_plane:
+            if not self._verify_worker_version():
+                self.set_unit_failed(
+                    cause="The k8s worker charm version does not meet the requirements."
```
eaudetcobello, line 185: I'm a bit confused as to why verify_worker_version is singular but internally it calls get_worker_versions which is plural.

eaudetcobello (commented):
Nice work! Left a few comments.

charms/worker/k8s/src/snap.py
mateoflorido: This is not the case here. If the refresh is blocked (due to a different revision) but the upgrade machinery has verified and allowed the upgrade, using an `or` in this context will cause the block to execute every time, since one of the conditions (`block_refresh`) will always evaluate to true.

mateoflorido: @addyess Sounds good. I'll push the check into the block_refresh method.

charms/worker/k8s/src/upgrade.py
mateoflorido: Originally, the method verified the consensus version for the workers. But now, we have changed that in order to accommodate multiple `k8s-worker` applications in the model. Context: https://github.com/canonical/k8s-operator/pull/200#discussion_r1866149071

charms/worker/k8s/src/charm.py
```diff
@@ -227,6 +231,35 @@ def is_worker(self) -> bool:
...
+
+        versions = defaultdict(list)
+        for unit in relation.units:
+            if version := relation.data[unit].get("version"):
+                versions[version].append(unit)
+        return versions
+
+    def grant_upgrade(self):
+        """Grant the upgrade to the charm."""
+        self._upgrade_snap = True
```
eaudetcobello, line 256: Not used

charms/worker/k8s/src/charm.py
```diff
@@ -227,6 +231,35 @@ def is_worker(self) -> bool:
...
+
+    @property
+    def is_upgrade_granted(self) -> bool:
+        """Check if the upgrade has been granted."""
+        return self._upgrade_snap
```
eaudetcobello, line 261: Not used

charms/worker/k8s/src/charm.py
eaudetcobello: Nvm, just saw it's used.

charms/worker/k8s/src/charm.py
eaudetcobello: Nvm, just saw it's used.

charms/worker/k8s/src/charm.py
mateoflorido: Used here: https://github.com/canonical/k8s-operator/blob/7e7e480eb5a5fe53f366f26231fe8995d7f39d0c/charms/worker/k8s/src/upgrade.py#L199

charms/worker/k8s/src/charm.py
mateoflorido: Used here: https://github.com/canonical/k8s-operator/blob/7e7e480eb5a5fe53f366f26231fe8995d7f39d0c/charms/worker/k8s/src/snap.py#L276

eaudetcobello (approved):
Thanks!

PR #203, opened 2024-12-03
Pin Runner Size/Arch

mateoflorido:


addyess (approved):
Fine... new day, new labels

.github/workflows/integration_test.yaml
```diff
@@ -39,8 +39,8 @@ jobs:
     strategy:
       matrix:
         arch:
-        - {id: amd64, builder-label: ubuntu-22.04, tester-arch: x64}
-        - {id: arm64, builder-label: ARM64,        tester-arch: ARM64}
+        - {id: amd64, builder-label: jammy, tester-arch: AMD64, size: large}
+        - {id: arm64, builder-label: jammy, tester-arch: ARM64, size: large}
```
addyess, line 43: Actually not fine... you can't switch the `builder-label` to just `jammy` b/c the self-hosted runners have labels `jammy` for both amd64 and arm64 machines.   You could end up building for one arch  and trying to "test" on another...

PR #217, opened 2024-12-14
Adjustments to test charm upgrades

addyess:
### Overview

Create tests that performs a charm upgrade from k8s edge --> k8s local_path

### Details
* eliminates `k8s-worker` leader from checking node readiness on `pre-upgrade-check` action
* `k8s` leader is responsible for checking the whole cluster on `pre-upgrade-check`
* `k8s` checks `kube-system` namespace to ensure all pods are ready during update-status event
* heavy refactor of integration test `conftest.py` to support test configuration by pytest model markings
* `latest/edge` snap is now 1.32, requiring new dependency literals
* increase runner size for amd64 runs

>  [!NOTE]
>  VMs on LXD cannot pull an image for focal that supports cilium.  This PR moves all testing on amd64 therefore to jammy rather than focal.  Focal is still tested on arm64 due to GH ARM64 workers not supporting `/dev/kvm`

### DIFF
--- a/charms/worker/k8s/src/inspector.py
+++ b/charms/worker/k8s/src/inspector.py
@@ -44,7 +44,7 @@ class ClusterInspector:
             self.client = Client(config=config.get())
         return self.client
 
-    def get_nodes(self, labels: LabelSelector) -> Optional[List[Node]]:
+    def get_nodes(self, labels: Optional[LabelSelector] = None) -> Optional[List[Node]]:
         """Get nodes from the cluster.
 
         Args:
@@ -56,6 +56,7 @@ class ClusterInspector:
         Raises:
             ClusterInspectorError: If the nodes cannot be retrieved.
         """
+        labels = labels or {}
         client = self._get_client()
         try:
 
--- a/charms/worker/k8s/tests/unit/test_upgrade.py
+++ b/charms/worker/k8s/tests/unit/test_upgrade.py
@@ -21,6 +21,7 @@ class TestK8sUpgrade(unittest.TestCase):
     def setUp(self):
         """Set up common test fixtures."""
         self.charm = MagicMock()
+        self.charm.is_worker = False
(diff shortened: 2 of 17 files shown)


charms/worker/k8s/src/inspector.py
```diff
@@ -56,6 +56,7 @@ def get_nodes(self, labels: LabelSelector) -> Optional[List[Node]]:
         Raises:
             ClusterInspectorError: If the nodes cannot be retrieved.
         """
+        labels = labels or {}
```
addyess, line 59: labels are now "OPTIONAL"

charms/worker/k8s/tests/unit/test_upgrade.py
```diff
@@ -64,15 +65,14 @@ def test_pre_upgrade_check_control_plane_success(self):
 
         self.upgrade.pre_upgrade_check()
 
-        self.node_manager.get_nodes.assert_called_once_with(labels={"juju-charm": "k8s"})
+        self.node_manager.get_nodes.assert_called_once_with()
```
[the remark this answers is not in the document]
addyess: @mateoflorido I think i'd rather make sure that `get_nodes` is called specifically with no arguments in this case.  Do you disagree?

tests/integration/helpers.py
```diff
@@ -80,22 +87,29 @@ async def get_unit_cidrs(model: Model, app_name: str, unit_num: int) -> List[str
...
 
     Args:
         k8s: any k8s unit
+        resource: string resource type
+        namespace: string namespace
+        labels: dict of labels
 
     Returns:
-        list of nodes
+        list of pods
```
mateoflorido, line 100: nit: Improve the docstring. The method retrieves any k8s resource and returns a list of the specified resource.

mateoflorido (approved):
Great job! Thanks for adding testing for the upgrades and improving the stability of the suite.

PR #223, opened 2024-12-18
Pin snap_installations to 1.32-classic stables

addyess:
### Overview
Pin the revisions based on `snapcraft status k8s`

```
1.32-classic  amd64   stable            v1.32.0       1843        -           -
                      candidate         v1.32.0       1843        -           -
                      beta              v1.32.0       1848        -           -
                      edge              v1.32.0       1848        -           -
              arm64   stable            v1.32.0       1844        -           -
                      candidate         v1.32.0       1849        -           -
                      beta              v1.32.0       1871        -           -
                      edge              v1.32.0       1871        -           -
```

### Details
* Quite a bit of work here to improve charm upgrades when no suitable source channels exist to upgrade from.
* Removal of the snap installation override feature b/c upgrading all control plane nodes simultaneously doesn't go well
* ensure we run the pre-upgrade-check on workers before upgrading

### DIFF
--- a/charms/worker/k8s/src/snap.py
+++ b/charms/worker/k8s/src/snap.py
@@ -279,10 +279,9 @@ def management(charm: K8sCharmProtocol) -> None:
         if isinstance(args, SnapFileArgument) and which.revision != "x1":
             snap_lib.install_local(**install_args)
         elif isinstance(args, SnapStoreArgument) and args.revision:
-            if which.revision != args.revision:
-                log.info("Ensuring %s snap revision=%s", args.name, args.revision)
-                which.ensure(**install_args)
-                which.hold()
+            log.info("Ensuring %s snap revision=%s", args.name, args.revision)
+            which.ensure(**install_args)
+            which.hold()
         elif isinstance(args, SnapStoreArgument):
             log.info("Ensuring %s snap channel=%s", args.name, args.channel)
             which.ensure(**install_args)
--- a/charms/worker/k8s/src/snap.py
+++ b/charms/worker/k8s/src/snap.py
@@ -110,6 +110,15 @@ class SnapStoreArgument(BaseModel):
             raise ValueError(f"Revision is not an integer: {value}")
         return value
 
+    def __str__(self):
+        """Represent the snap store argument as a string.
+
+        Returns:
+            str: The string representation
+        """
+        _type = type(self).__name__
+        return f"<{_type}: {self.name}-{self.revision}.{self.channel} -- {self.state}>"
+
 
 SnapArgument = Annotated[
     Union[SnapFileArgument, SnapStoreArgument], Field(discriminator="install_type")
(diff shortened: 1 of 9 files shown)


charms/worker/k8s/src/snap.py
```diff
@@ -279,10 +279,9 @@ def management(charm: K8sCharmProtocol) -> None:
         if isinstance(args, SnapFileArgument) and which.revision != "x1":
             snap_lib.install_local(**install_args)
         elif isinstance(args, SnapStoreArgument) and args.revision:
-            if which.revision != args.revision:
-                log.info("Ensuring %s snap revision=%s", args.name, args.revision)
-                which.ensure(**install_args)
-                which.hold()
+            log.info("Ensuring %s snap revision=%s", args.name, args.revision)
+            which.ensure(**install_args)
+            which.hold()
```
addyess, line 284: @mateoflorido i think we need to talk about this.  Sorry to call you out to look at it. 

Notice this PR is into the new release_1.32 branch.  I'll open another PR into main.  

but I think `block_refresh` has already checked that the revision isn't changing so it's not necessary to do that here.  I ran into an issue because the `which.revision` was the correct snap revision but the state of the snap was "Available" and not present. 

I can change this to 


```suggestion
            if which.revision != args.revision and which.state != snap_lib.SnapState.Present:
                log.info("Ensuring %s snap revision=%s", args.name, args.revision)
                which.ensure(**install_args)
                which.hold()
```

addyess: also the snap library `ensure(...)` prevents a refresh if we're on the right revision

charms/worker/k8s/src/snap.py
```diff
@@ -269,23 +283,28 @@ def management(charm: K8sCharmProtocol) -> None:
...
             snap_lib.install_local(**install_args)
-        elif isinstance(args, SnapStoreArgument) and args.revision:
-            if which.revision != args.revision:
-                log.info("Ensuring %s snap revision=%s", args.name, args.revision)
+        elif isinstance(args, SnapStoreArgument):
+            log.info("Ensuring args=%s current=%s", str(args), str(which))
+            new_rev = bool(args.revision) and which.revision != args.revision
+            new_channel = bool(args.channel) and which.channel != args.channel
+            new_state = which.state not in (snap_lib.SnapState.Present, snap_lib.SnapState.Latest)
+            if any([new_rev, new_channel, new_state]):
```
addyess, line 303: I think these changes collapse nicely.  we run `ensure and hold` always regardless of whether we're following a revision or a channel.  Cluster wide upgrades are ill-advised if they are done at the same moment

tests/integration/test_upgrade.py
```diff
@@ -16,20 +17,53 @@
...
+            out = subprocess.check_output(
+                ["juju", "info", app, "--channel", f"{track}/{lookup}", "--format", "yaml"]
+            )
+            track_map = yaml.safe_load(out).get("channels", {}).get(track, {})
+            if lookup in track_map:
+                log.info("Found %s in %s", app, f"{track}/{lookup}")
+                break
+        else:
+            return True, f"No suitable channel found for {app} in {channel} to upgrade from"
+    return False, ""
```
addyess, line 51: There are cases (like this one) where there's nothing to upgrade FROM.  I had to artificially place something in the 1.32/beta branch just so that tests have something to upgrade from.  

I certainly like this "skipif" feature on the module.  It may make sense in the future to run through multiple upgrades

test-1: upgrade to THESE changes from the last stable release 
test-2: upgrade to THESE changes from the last beta on this channel

tests/integration/test_upgrade.py
```diff
@@ -73,10 +107,8 @@ async def _refresh(app_name: str):
         action = await leader.run_action("pre-upgrade-check")
         await action.wait()
         with_fault = f"Pre-upgrade of '{app_name}' failed with {yaml.safe_dump(action.results)}"
-        if app_name == "k8s":
-            # The k8s charm has a pre-upgrade-check action that works, k8s-worker does not.
-            assert action.status == "completed", with_fault
-            assert action.results["return-code"] == 0, with_fault
+        assert action.status == "completed", with_fault
+        assert action.results["return-code"] == 0, with_fault
```
addyess, line 111: oopsie, we must confirm the pre-upgrade check for both k8s and k8s-worker now

charms/worker/k8s/src/snap.py
```diff
@@ -269,23 +283,28 @@ def management(charm: K8sCharmProtocol) -> None:
...
         if isinstance(args, SnapFileArgument) and which.revision != "x1":
             snap_lib.install_local(**install_args)
-        elif isinstance(args, SnapStoreArgument) and args.revision:
-            if which.revision != args.revision:
-                log.info("Ensuring %s snap revision=%s", args.name, args.revision)
+        elif isinstance(args, SnapStoreArgument):
+            log.info("Ensuring args=%s current=%s", str(args), str(which))
+            new_rev = bool(args.revision) and which.revision != args.revision
+            new_channel = bool(args.channel) and which.channel != args.channel
+            new_state = which.state not in (snap_lib.SnapState.Present, snap_lib.SnapState.Latest)
```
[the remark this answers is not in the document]
mateoflorido: OHHH. Nevermind, I misunderstood `302`. So if it's present, we should just carry on.

mateoflorido (approved):
LGTM!

PR #226, opened 2024-12-19
Provide tag-prefix when promoting charms

addyess:
### Overview
* multi charm repos like this one have tags per charm and per revision.  such as `k8s-rev123`. 
* The operator-workflows should pass this `tag-prefix` along during the promote job

Depends on 
* https://github.com/canonical/operator-workflows/pull/508


louiseschmidtgen (approved):
Nice work!

PR #234, opened 2025-01-06
Add external-load-balancer relation

HomayoonAlimohammadi:
### Overview
This PR adds `external-load-balancer` relation. This relation enables `k8s-operator` to utilize external LB sources like Openstack Octavia or Amazon ELB.

The `external-load-balancer-port` config is also added. This config should be set according to the external LB provider config. Example for Openstack integrator: https://github.com/charmed-kubernetes/charm-openstack-integrator/blob/main/reactive/openstack.py#L245-L254

Please merge me as well https://github.com/canonical/k8s-operator/pull/235 (Happy new year!)


charms/worker/k8s/src/charm.py
```diff
@@ -383,6 +397,64 @@ def _assemble_bootstrap_config(self):
...
+                log.error("Error from external load balancer: %s", res.error_message)
+                raise ReconcilerError(
+                    "Failed to configure external load balancer. Check logs for details"
+                )
+
+            log.info("External load balancer response received: %s", res)
+            return
+
+        log.error(
+            "Timedout waiting for external load balancer response after %s seconds",
```
bschimke95, line 453: nit: typo

charms/worker/k8s/charmcraft.yaml
```diff
@@ -389,6 +389,12 @@ config:
       description: |
         Labels can be used to organize and to select subsets of nodes in the
         cluster. Declare node labels in key=value format, separated by spaces.      
+    external-load-balancer-port:
```
bschimke95, line 392: Is this the recommended approach to configure such a port?
I feel like this port should be configured by/on the load-balancer charm and then shared through the `external-load-balancer` relation but I'm happy to learn something if this is not the case.

bschimke95 (commented):
great work @HomayoonAlimohammadi 
Did a first pass with a general question.

charms/worker/k8s/src/charm.py
```diff
@@ -362,9 +373,13 @@ def _check_k8sd_ready(self):
...
-        configured_sans = {san for san in extra_sans_str.strip().split() if san}
-        all_sans = configured_sans | set([_get_public_address()])
-        return sorted(all_sans)
+        extra_sans = {san for san in extra_sans_str.strip().split() if san}
+        if public_address := self._get_public_address():
+            log.info("Public address %s found, adding it to extra SANs", public_address)
+            extra_sans.add(public_address)
+        else:
+            log.info("No public address found, skipping adding public address to extra SANs")
+        return sorted(extra_sans)
```
addyess, line 382: @HomayoonAlimohammadi won't the public address from the loadbalancer need to be added to the sans?

charms/worker/k8s/src/charm.py
```diff
@@ -383,6 +398,64 @@ def _assemble_bootstrap_config(self):
...
+                )
+
+            log.info("External load balancer response received: %s", res)
+            return
+
+        log.error(
+            "Timed out waiting for external load balancer response after %s seconds",
+            retry_interval_seconds * retries,
+        )
+        raise ReconcilerError("External load balancer response not received")
```
addyess, line 457: Alright, so the response can't be "waited on".  juju charms are event based.  You're going to make a request in one event say (relation-joined) and you'll see the data available to use later on another event `relation-changed`.  This retry/wait loops therefore fruitless.

let's instead just try to get it, and if it's not there -- raise your error.  Later on, juju will give you another event to let you know the relation has changed and the response will be there.

charms/worker/k8s/src/charm.py
```diff
@@ -383,6 +398,64 @@ def _assemble_bootstrap_config(self):
         config.extra_args.craft(self.config, bootstrap_config, cluster_name)
         return bootstrap_config
 
+    @on_error(ops.WaitingStatus("Waiting for external load balancer"), ReconcilerError)
+    def _configure_external_load_balancer(self):
```
addyess, line 402: I think you'd be good to consult the following 
* https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/main/src/charm.py#L388

I'd think you could lift half of the methods in it for the lb_external

charms/worker/k8s/src/charm.py
```diff
@@ -1059,14 +1133,81 @@ def _get_external_kubeconfig(self, event: ops.ActionEvent):
...
+        """
+        addr = addr.lower()
+        if not addr.startswith(HTTPS_SCHEME) and not addr.startswith(HTTP_SCHEME):
+            log.info("Adding %s to addr %s", HTTPS_SCHEME, addr)
+            addr = HTTPS_SCHEME + addr
+        if re.search(ENDPOINT_HAS_PORT_REGEX, addr) is None:
+            log.info("Adding port %s to addr %s", port, addr)
+            addr = f"{addr}:{port}"
+        return addr
+
```
addyess, line 1210: You may like [this module](https://github.com/charmed-kubernetes/charm-kubernetes-control-plane/blob/main/src/k8s_api_endpoints.py)

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: Honestly that's what I exactly tried to do. Do you spot any discrepancy or issues between the two?

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: I added a commit that removes this part and I also noticed that it's probably useless to check for the response here. 
We only care about this response when we're trying to get the public address. We can raise errors and fail then.

addyess: precisely

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: Honestly this module is again exactly what I've been trying to replicate. I just tried to be a bit more cautious in this PR (making sure that we don't have a port or scheme before adding one).

HomayoonAlimohammadi: No it's flawed... Lemme have a look at it tomorrow. I need to make sure we consider every scenario to prevent building invalid URLs, e.g. with or without scheme, ipv4 or ipv6, with or without port, ipv6 with or without braces around, ... .

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: I'm not sure if I'm getting the problem here. We're trying to add the public address to the SANs by `self._get_public_address()`. Won't this include that address in the SANs?

charms/worker/k8s/src/charm.py
```diff
@@ -361,11 +369,19 @@ def _check_k8sd_ready(self):
...
+        """Retrieve the certificate extra SANs.
+
+        Raises:
+            ReconcilerError: If the public address cannot be retrieved.
+        """
         extra_sans_str = str(self.config.get("kube-apiserver-extra-sans") or "")
-        configured_sans = {san for san in extra_sans_str.strip().split() if san}
-        all_sans = configured_sans | set([_get_public_address()])
-        return sorted(all_sans)
+        extra_sans = {san for san in extra_sans_str.strip().split() if san}
```
addyess, line 378: Funny, I don't think pythons's `str.split()` can ever return an empty string.  I'm not sure if the filter checking is even necessary

```suggestion
        extra_sans = set(extra_sans_str.strip().split())
```

charms/worker/k8s/src/charm.py
```diff
@@ -361,11 +369,19 @@ def _check_k8sd_ready(self):
...
+        if public_address := self._get_public_address():
+            log.info("Public address %s found, adding it to extra SANs", public_address)
+            extra_sans.add(public_address)
+        else:
+            raise ReconcilerError("Failed to get public address")
```
addyess, line 383: honestly, i don't get it... when are we ever going to fail to get a pubilc address?  If we don't get one -- let's just leave it out of the sans list -- why are we failing the reconciler?

charms/worker/k8s/src/charm.py
```diff
@@ -361,11 +369,19 @@ def _check_k8sd_ready(self):
...
+        return sorted(extra_sans)
```
addyess, line 384: ⚠️  We need the response from the `external_load_balancer` to be added to the SANS.  if there's not this relation or there's no response -- leave it out of the set list

addyess: we ALSO need the public address of each of the control-plane units.  ALL of these should be in the sans.

k8s/0 - ingress-address from the cluster relation
k8s/1 - ingress-address from the cluster relation
k8s/2 - ingress-address from the cluster relation
loadbalancer-ip
all charm configured extra sans

ALL these IPs should be in the sans once -- don't exclude the unit ips if there is a load balancer

charms/worker/k8s/src/charm.py
```diff
@@ -1061,14 +1106,62 @@ def _get_external_kubeconfig(self, event: ops.ActionEvent):
...
+
+        Returns:
+            str: public ip address of the unit
+            None: if the public address is not available
+        """
+        if self.is_control_plane and self.external_load_balancer.is_available:
+            log.info("Using external load balancer address as the public address")
+            return self._get_external_load_balancer_address()
+        log.info("Using juju public address as the public address")
+        return _get_juju_public_address()
```
addyess, line 1138: maybe a better comment for this function is

>  the "most public" ip address.  But don't ignore the unit's IP in the sans just because there's a load balancer

charms/worker/k8s/src/endpoints.py
```diff
@@ -0,0 +1,92 @@
...
+            # either ipv6 without braces or ipv4+port
+            if netloc.count(":") > 1:
+                # ipv6 without braces and without port.
+                # an ipv6 without braces but with port is technically indiscriminable
+                # from another ipv6 without port so we don't consider it.
+                is_ipv6 = True
+                ip = netloc
+            else:
+                # ipv4+port
+                ip, port = netloc.split(":")
```
addyess, line 59: I think urlparse handles this.  `You can use netloc.port` to find the port if the port is available:

```python
>>> from urllib.parse import urlparse
>>> urlparse("http://1.2.3.4:80").netloc
1.2.3.4:80
>>> urlparse("http://1.2.3.4:80").port
80
>>> urlparse("http://1.2.3.4:80").hostname
1.2.3.4
```

addyess: if the port is unspecified it will be `None`

charms/worker/k8s/src/endpoints.py
```diff
@@ -0,0 +1,92 @@
...
+    scheme, ip, port, is_ipv6 = parse_endpoint(addr)
+
+    if scheme:
+        log.info(f"replacing already available scheme {scheme} in {addr=} with {new_scheme}")
+    if port:
+        log.info(f"replacing already available port {port} in {addr=} with {new_port}")
+    if is_ipv6:
+        ip = f"[{ip}]"
+
+    return f"{new_scheme}://{ip}:{new_port}"
```
addyess, line 92: this looks an awful lot like [`urlunsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunsplit)

charms/worker/k8s/src/charm.py
```diff
@@ -1061,14 +1106,62 @@ def _get_external_kubeconfig(self, event: ops.ActionEvent):
...
+                log.info("No server requested, use public address")
+                server = self._get_public_address()
+                if not server:
+                    event.fail("Failed to get public address. Check logs for details.")
+                    return
+                log.info("Found public address: %s", server)
+                port = str(APISERVER_PORT)
+                if self.is_control_plane and self.external_load_balancer.is_available:
+                    log.info("Using external load balancer port as the public port")
+                    port = str(EXTERNAL_LOAD_BALANCER_PORT)
```
addyess, line 1118: it feels like this should be a new method called `_get_public_endpoint()`.  It should return the "most public" endpoint available for the charm.  the order should be:

```
https://<external_load_balancer-ip>:LB_PORT
https://<leader CP unit ingress address>:INTERNAL_PORT
https://<other CP units ingress addresses>:INTERNAL_PORT
https://127.0.0.1:INTERNAL_PORT
```

it should ALWAYS arrive at SOMETHING if k8s is bootstrapped.

addyess (changes requested):
Awesome getting closer. 

the big take away is that the sans list should have EVERY address someone could use to try to get to this cluster -- not just the lead control plane, not just the load balancer.  Add them all.  That way the cert is valid for any entry. 

Secondarily, we'll need to refresh the sans list in the reconciler loop. Maybe there's some way to only refresh if there are changes.  But avoid using stored state.  If you can read the sans list FROM the existing certificate, that'd be better

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: So this is the code for getting the public address from an external load balancer:
```python
    def _get_external_load_balancer_address(self) -> Optional[str]:
        ...
        response = self.external_load_balancer.get_response(EXTERNAL_LOAD_BALANCER_RESPONSE_NAME)

        if not response:
            log.error("No response from external load balancer when trying to get address")
            return None
        if response.error:
            log.error(
                "Error from external load balancer when trying to get address: %s",
                response.error,
            )
            return None

        return response.address
```
This is the only place that we care about the loadbalancer provider response. I thought maybe we should fail if there's an error in that response or...

HomayoonAlimohammadi: Changing this part to only log something instead of failing the reconciliation.

charms/worker/k8s/src/endpoints.py
HomayoonAlimohammadi: It honestly has _some_ corner cases. It doesn't handle ipv6 correctly. As an example checkout this input, output pair:
```
input = "2001:db8::1"
...
print(parsed.netloc, "--", parsed.hostname, parsed.port)
--> ValueError: Port could not be cast to integer value as 'db8::1'
```
And then it doesn't handle braces in the address very well. Some of these might not be cases we're interested in right now for public addresses, but IMO it's better be safe than sorry. WDYT?

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: > it should ALWAYS arrive at SOMETHING if k8s is bootstrapped.

That's right but at the same time if we have an external load balancer and we fail to retrieve its address, I think it's best to fail rather than returning the second best option (which is probably not so useful, we have a loadbalancer and we're probably don't want to talk directly to our backends).

charms/worker/k8s/requirements.txt
```diff
@@ -17,3 +17,5 @@ websocket-client==1.8.0
 poetry-core==1.9.1
 lightkube==0.16.0
 httpx==0.27.2
+loadbalancer_interface == 1.2.0
+cryptography==44.0.0
```
addyess, line 21: OH dude, i have bad news.  You can't use this library b/c that library requires our charm not be able to support multiple series.  

`cryptography` compiles C code specific for the ubuntu base's python.  Building on focal, yields code that doesn't run on jammy or noble.  The same is true if you build on jammy -- it doesn't run on focal/noble. 

my best advice.  subprocess out to `openssl` on the device

charms/worker/k8s/src/charm.py
```diff
@@ -362,10 +371,29 @@ def _check_k8sd_ready(self):
...
+        # Add the ingress addresses of all units
+        extra_sans.add(_get_juju_public_address())
+        binding = self.model.get_binding(CLUSTER_RELATION)
+        try:
+            addresses = binding and binding.network.ingress_addresses
+            if addresses:
+                for addr in addresses:
+                    extra_sans.add(str(addr))
+        except ops.RelationNotFoundError as e:
+            log.error(f"Failed to get ingress addresses for extra SANs: {e}")
```
addyess, line 387: so, the CLUSTER_RELATION should ALWAYS be available b/c it's a peer relation.  it's required to be there 

Would this work?
```suggestion
        binding = self.model.get_binding(CLUSTER_RELATION)
        extra_sans |= {str(addr) for addr in binding and binding.network.ingress_addresses or [])
```

charms/worker/k8s/src/charm.py
```diff
@@ -362,10 +371,29 @@ def _check_k8sd_ready(self):
...
+
+        # Add the external load balancer address
+        if self.is_control_plane and self.external_load_balancer.is_available:
```
addyess, line 390: it doesn't make sense for the worker to need this list -- basically ever. If you wanna protect against it occuring, just do a this near the beginning of the func

```python
if not self.is_control_plane:
    raise ReconcilerError("Intended only for the Control Plane Charm")
```

charms/worker/k8s/src/pki.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
```
addyess, line 1: meh -- we aren't running this file.  Just remove this line.

charms/worker/k8s/src/pki.py
```diff
@@ -0,0 +1,67 @@
...
+"""A module providing PKI related functionalities."""
+
+import ipaddress
+import logging
+import os
+import typing
+
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend
+from cryptography.x509.extensions import ExtensionNotFound
```
addyess, line 17: sigh... we can't use this :( as it depends on libraries that are series specific.  

grow to love `subprocess.check_output()`

addyess (changes requested):
Dude, hue this is looking so good.  Feel free to be angry with me.
Thanks for your defensive coding practices

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: This was because of some unit tests that were expecting this relation but it was not there.

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: Tried removing this but we're calling `_assemble_bootstrap_config` from a unit test with both a control plane and a worker charm.

charms/worker/k8s/src/charm.py
```diff
@@ -1061,14 +1120,102 @@ def _get_external_kubeconfig(self, event: ops.ActionEvent):
...
+        """Get the most public address either from external load balancer or from juju.
+
+        If the external load balancer is available and the unit is a control-plane unit,
+        the external load balancer address will be used. Otherwise, the juju public address
+        will be used.
+        NOTE: Don't ignore the unit's IP in the extra SANs just because there's a load balancer.
+
+        Returns:
+            str: public ip address of the unit
+            None: if the public address is not available
```
mateoflorido, line 1154:
```suggestion
        Returns:
            Optional[str]: The public IP address of the unit, or None if the public address is not available.
```

charms/worker/k8s/src/charm.py
```diff
@@ -1061,14 +1120,102 @@ def _get_external_kubeconfig(self, event: ops.ActionEvent):
...
+            return self._get_external_load_balancer_address()
+        log.info("Using juju public address as the public address")
+        return _get_juju_public_address()
+
+    def _get_external_load_balancer_address(self) -> Optional[str]:
+        """Get the external load balancer address.
+
+        Returns:
+            None: If the external load balancer address is not available.
+            str: The external load balancer address.
```
mateoflorido, line 1167:
```suggestion
        Returns:
            Optional[str]: The external load balancer IP address, or None if the address is not available.
```

charms/worker/k8s/src/charm.py
```diff
@@ -1061,14 +1120,102 @@ def _get_external_kubeconfig(self, event: ops.ActionEvent):
...
+        """
+        if not self.is_control_plane:
+            log.error("External load balancer address is only available for control-plane units")
+            return None
+
+        response = self.external_load_balancer.get_response(EXTERNAL_LOAD_BALANCER_RESPONSE_NAME)
+
+        if not response:
+            log.error("No response from external load balancer when trying to get address")
+            return None
```
mateoflorido, line 1177: This might be a transient error. What if we adjust the log level here?
```suggestion
        if not response:
            log.info("Response from external load balancer is not yet available.")
            return None
```

charms/worker/k8s/src/charm.py
```diff
@@ -1061,14 +1120,102 @@ def _get_external_kubeconfig(self, event: ops.ActionEvent):
...
+        dns_sans, ip_addresses = get_certificate_sans(APISERVER_CERT)
+        ip_addresses = [str(ip) for ip in ip_addresses]
+        all_cert_sans = dns_sans + ip_addresses
+
+        for san in extra_sans:
+            if san not in all_cert_sans:
+                log.info(f"{san} not in cert SANs, refreshing certs with new SANs: {extra_sans}")
+                status.add(ops.MaintenanceStatus("Refreshing Certificates"))
+                self.api_manager.refresh_certs(extra_sans)
+                log.info("Certificates have been refreshed")
```
mateoflorido, line 1215: This calls `refresh_certs` in a loop, once for each missing SAN. Therefore, the certificates will be refreshed multiple times unnecessarily:
```suggestion
        missing_sans = [san for san in extra_sans if san not in all_cert_sans]
        if missing_sans:
                log.info("Missing SANs found, refreshing certificates.")
                status.add(ops.MaintenanceStatus("Refreshing Certificates"))
                self.api_manager.refresh_certs(extra_sans)
                log.info("Certificates have been refreshed")
```

pyproject.toml
```diff
@@ -48,7 +48,8 @@ plugins = "pydantic.mypy"
 [tool.pylint]
 # Ignore too-few-public-methods due to pydantic models
 # Ignore no-self-argument due to pydantic validators
-disable = "wrong-import-order,redefined-outer-name,too-many-instance-attributes,too-few-public-methods,no-self-argument,fixme,protected-access"
+# Ignore logging-fstring-interpolation which instructs to use %s instead of f-strings in logs which is objectively worse.
```
mateoflorido, line 51: F-strings are evaluated even if the log message is not displayed (e.g., log level): https://docs.python.org/3/howto/logging.html#optimization

mateoflorido (changes requested):

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: That's a really good catch. Thanks!

pyproject.toml
HomayoonAlimohammadi: Nice, TIL! Changing to %s, but TBH I'm not fully convinced that this optimization is actually worth the readability hit.

HomayoonAlimohammadi:
> I support Adam's suggestion. You can resolve the unit testing issue by using the Harness and creating the required relations, right?

@mateoflorido Since the test is basically testing `_assemble_bootstrap_config` I think it's best to skip it for the worker charms. WDYT?
/me changes the test and the `except`...

charms/worker/k8s/src/charm.py
```diff
@@ -385,6 +410,34 @@ def _assemble_bootstrap_config(self):
...
+            return
+
+        status.add(ops.MaintenanceStatus("Configuring external loadBalancer"))
+
+        req = self.external_load_balancer.get_request(EXTERNAL_LOAD_BALANCER_REQUEST_NAME)
+        req.protocol = req.protocols.tcp
+        req.port_mapping = {EXTERNAL_LOAD_BALANCER_PORT: APISERVER_PORT}
+        req.public = True
+        if not req.health_checks:
+            req.add_health_check(protocol=req.protocols.http, port=APISERVER_PORT, path="/livez")
```
addyess, line 437: Hmm?  shouldn't this be https not http?

Also, I have checked this endpoint `/livez` is disabled because of the user of `--anonymous-auth=false` in the kube-apiserver.  I think we'll require users config the `kube-apiserver-extra-args=anonymous-auth=true` to expose this health check. 


The HealthCheck doesn't have any way to do authentication so this flag will need to be enabled for the load-balancer to accurately do its job -- otherwise it will think the API server is always "down"

charms/worker/k8s/src/charm.py
```diff
@@ -1061,14 +1116,104 @@ def _get_external_kubeconfig(self, event: ops.ActionEvent):
...
+
+        if not response:
+            log.info("Response from external load balancer is not yet available.")
+            return None
+        if response.error:
+            log.error(
+                "Error from external load balancer when trying to get address: %s",
+                response.error,
+            )
+            return None
```
addyess, line 1179: i think if we get an error response from the load-balancer, we likely want to expose this to the user through a blocked state or some kind.  the integrator charm providing the load-balancer couldn't make a load-balancer because there is something wrong with the request.  Usually this means the juju admin needs to take action to resolve it.  In those cases we normally "block"

can we raise an `ReconcilerError(f"Load Balancer: {response.error}")`

it's  better to check the response in the `_configure_external_load_balancer` method,  

in there we can check if there is a response, 
* if no response, we are "waiting on load-balancer"
* if there is a response and its an error we are "blocked on load-balancer issue". 

this should make this method much smaller b/c there's less check... check... checking going on.

WDYT?

addyess (changes requested):
Looking really really good.  Thanks for the switch to openssl. 

I'm going to build this charm today and try it out

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -920,3 +981,34 @@ def get_kubeconfig(self, server: Optional[str]) -> str:
...
+            expiration_seconds = int((twenty_years_later - now).total_seconds())
+
+        run_endpoint = "/1.0/k8sd/refresh-certs/run"
+        run_req = RefreshCertificatesRunRequest(  # type: ignore
+            seed=plan_resp.metadata.seed,
+            expiration_seconds=expiration_seconds,
+            extra_sans=extra_sans,
+        )
+
+        body = run_req.dict(exclude_none=True, by_alias=True)
```
addyess, line 1013: i'm seeing this failure when testing:

```
Request failed:
        method=POST
        endpoint=/1.0/k8sd/refresh-certs/plan
2025-01-24 04:16:19 ERROR unit.k8s/0.juju-log server.go:326 Caught ReconcilerError
  File "/var/lib/juju/agents/unit-k8s-0/charm/lib/charms/k8s/v0/k8sd_api_manager.py", line 815, in _send_request
    return response_cls.parse_raw(data)
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/pydantic/main.py", line 549, in parse_raw
    return cls.parse_obj(obj)
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/pydantic/main.py", line 526, in parse_obj
    return cls(**obj)
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/pydantic/main.py", line 341, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for RefreshCertificatesPlanResponse
metadata -> seed
  field required (type=value_error.missing)

The above exception was the direct cause of the following exception:

  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/charms/contextual_status.py", line 101, in on_error
    yield
  File "/usr/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/var/lib/juju/agents/unit-k8s-0/charm/./src/charm.py", line 1236, in _ensure_cert_sans
    self.api_manager.refresh_certs(extra_sans)
  File "/var/lib/juju/agents/unit-k8s-0/charm/lib/charms/k8s/v0/k8sd_api_manager.py", line 995, in refresh_certs
    plan_resp = self._send_request(plan_endpoint, "POST", RefreshCertificatesPlanResponse, {})
  File "/var/lib/juju/agents/unit-k8s-0/charm/lib/charms/k8s/v0/k8sd_api_manager.py", line 818, in _send_request
    raise InvalidResponseError(
charms.k8s.v0.k8sd_api_manager.InvalidResponseError: Error status 200
Request failed:
        method=POST
        endpoint=/1.0/k8sd/refresh-certs/plan

The above exception was the direct cause of the following exception:

  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/charms/reconciler.py", line 35, in reconcile
    self.reconcile_function(event)
  File "/var/lib/juju/agents/unit-k8s-0/charm/./src/charm.py", line 991, in _reconcile
    self._ensure_cert_sans()
  File "/usr/lib/python3.10/contextlib.py", line 78, in inner
    with self._recreate_cm():
  File "/usr/lib/python3.10/contextlib.py", line 153, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/charms/contextual_status.py", line 106, in on_error
    raise ReconcilerError(msg) from e
charms.contextual_status.ReconcilerError: Found expected exception: Error status 200
Request failed:
        method=POST
        endpoint=/1.0/k8sd/refresh-certs/plan
```

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -598,6 +599,66 @@ class GetKubeConfigResponse(BaseRequestModel):
...
+class RefreshCertificatesRunRequest(BaseModel, allow_population_by_field_name=True):
+    """Request model for running the refresh certificates run.
+
+    Attributes:
+        seed (int): The seed for the new certificates from plan response.
+        expiration_seconds (int): The duration of the new certificates.
+        extra_sans (list[str]): List of extra sans for the new certificates.
+    """
+
+    seed: int
```
addyess, line 636: Found it -- the response json has "seconds" here. 

```suggestion
    seed: int = Field(default=None, alias="seconds")
```

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
addyess: resolution is [here](https://github.com/canonical/k8s-operator/pull/234/files#r1928084363)

addyess (approved):
Finally, it all looks so good.  We know it's not 100% tested yet but the impl looks right.  Nice job @HomayoonAlimohammadi

HomayoonAlimohammadi:
Thanks @addyess! I need to add multiple unit and integration tests. Didn't test the Openstack LB integration since you mentioned the issue with the openstack integrator charm (incompatible ops versions), but manually tried changing the extra SANs through charm config and seems like the certs refresh alright.

mateoflorido (approved):

PR #237, opened 2025-01-07
[Integration test] Allows for `clouds(...)` marker and `--no-deploy` handling

addyess:
### Overview
Facilitates running most of the tests on existing models that are already deployed handling situations where specific tests or entire tests modules can be executed.  Some tests will be excluded if they match the wrong cloud or require changing the existing model. 

* allows tests to be marked with `clouds(...)` in order to skip tests unless they are on a certain cloud
* allows tests to skip certain tests if the `--no-deploy` flag is provided and the tests would deploy new applications/units into the model

### Details
* Uses the `no-deploy` pytest variable from `pytest-operator` to skip deployment of applications into the main kubernetes model in the event this is a model intended to be reused
* Creates an auto-used fixture which confirms the cloud-type can run the tests if the test is marked with the `clouds` marker.
* replace the `abort_on_fail` marker with a simpler `exitfirst` argument to pytest.


bschimke95 (approved):
LGTM, I probably would have used test tags to mark tests for specific clouds but this also works.
Great work @addyess

PR #242, opened 2025-01-10
Require tf users to specify a charm channel

addyess:
### Overview

Don't default the charm channels to `1.30/edge` -- a channel which may be or is already closed. 
Instead, require the tf user to specify a channel

### Overview
* removes the default channel from the tf modules

### DIFF
--- a/charms/worker/k8s/terraform/variables.tf
+++ b/charms/worker/k8s/terraform/variables.tf
@@ -21,7 +21,6 @@ variable "base" {
 variable "channel" {
   description = "The channel to use when deploying a charm."
   type        = string
-  default     = "1.30/edge"
 }
 
 variable "config" {
--- a/charms/worker/terraform/variables.tf
+++ b/charms/worker/terraform/variables.tf
@@ -21,7 +21,6 @@ variable "base" {
 variable "channel" {
   description = "The channel to use when deploying a charm."
   type        = string
-  default     = "1.30/edge"
 }
 
 variable "config" {


louiseschmidtgen (commented):
Ideally we would have the default version be the latest stable (like Microk8s). I don't know whether that is possible with terraform?

addyess:
> Ideally we would have the default version be the latest stable (like Microk8s). I don't know whether that is possible with terraform?

The charms will never have a `latest/stable`.  I propose that each branch (release-1.32, 1.33, 1.34) reference that branch's stable channel.

louiseschmidtgen (commented):
Would it make sense for us to add a validation for the channel variable since we're dealing with user input here or does that create too much overhead?

charms/worker/k8s/terraform/variables.tf
```diff
@@ -21,7 +21,6 @@ variable "base" {
 variable "channel" {
   description = "The channel to use when deploying a charm."
   type        = string
-  default     = "1.30/edge"
```
HomayoonAlimohammadi, line 24: nit: Doesn't this have a negative impact on UX? Maybe changing this to `latest/stable` since IIRC this is `snap install k8s` does?

HomayoonAlimohammadi (approved):
Great stuff, thanks a lot @addyess! LGTM, just a minor question.

charms/worker/k8s/terraform/variables.tf
addyess: There will never be a `latest/stable` charm.  We can make `latest/edge` i guess.  But i'd rather not folks by default installing the edge version of the charm.  I'd rather people be FORCED to pick a version

louiseschmidtgen (approved):
Let's do this!

PR #255, opened 2025-01-17
Upgrade tests must continue when k8s leader is ready for worker upgrades

addyess:
### Overview

Handle charm upgrades that might upgrade to a different snap version (ie 1.32.0 -> 1.32.1)

### Details
The k8s (control-plane) charm will upgrade first, and signal with a "waiting" status that the workers will need to be updated. 

if the snap revision changes but doesn't bump the patch version, the k8s units will all report "active/idle"
if the snap revision changes but the patch version is different, the leader k8s unit will report "waiting/idle" with a specific message. 

Either of these 2 cases is enough of a gate to allow for upgrades to continue

### DIFF
--- a/tests/integration/test_upgrade.py
+++ b/tests/integration/test_upgrade.py
@@ -5,6 +5,7 @@
 
 """Upgrade Integration tests."""
 
+import datetime
 import logging
 import subprocess
 from typing import Iterable, Optional, Tuple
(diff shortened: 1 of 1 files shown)


tests/integration/test_upgrade.py
```diff
@@ -78,20 +79,48 @@ async def test_upgrade(kubernetes_cluster: juju.model.Model, ops_test: OpsTest):
...
+        stop=stop_after_attempt(10),
+        wait=wait_fixed(30),
+        before_sleep=before_sleep_log(log, logging.WARNING),
+    )
+    async def _wait_for_upgrade_complete():
+        """Wait for the model to become idle."""
+        k8s_apps = {
+            k: v
+            for k, v in kubernetes_cluster.applications.items()
+            if v.startswith(CONTROL_PLANE_APP)
```
mateoflorido, line 106: Shouldn't this be:
```suggestion
            if k.startswith(CONTROL_PLANE_APP)
```
IIRC, `applications` returns `app_name, Application`

mateoflorido (approved):

Issue #204, opened 2024-12-04
dmidecode is missing from host


gboutry:
### Bug Description

A sunbeam user encountered this issue while trying to deploy using this operator. Dmidecode is not present by default in his environment.

### To Reproduce

1. juju add-machine --base ubuntu@22.04
2. juju wait-for machine 0 --query='status=="started"'
3. juju ssh 0 sudo apt remove --yes dmidecode
4. juju deploy k8s --channel 1.31/beta --to 0
5. juju wait-for unit k8s/0 --query='workload-status=="error" && agent-status=="idle"'

### Environment

1.31/beta on a 22.04 host, charm deployed by sunbeam.

### Relevant log output

```shell
unit-k8s-0: 10:21:51 INFO unit.k8s/0.juju-log Running legacy hooks/install.
unit-k8s-0: 10:21:51 ERROR unit.k8s/0.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-k8s-0/charm/./src/charm.py", line 996, in <module>
    ops.main(K8sCharm)
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/ops/__init__.py", line 343, in __call__
    return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/ops/_main.py", line 541, in main
    manager = _Manager(charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/ops/_main.py", line 424, in __init__
    self.charm = self._make_charm(self.framework, self.dispatcher)
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/ops/_main.py", line 427, in _make_charm
    charm = self._charm_class(framework)
  File "/var/lib/juju/agents/unit-k8s-0/charm/./src/charm.py", line 155, in __init__
    self.distributor = TokenDistributor(self, self.get_node_name(), self.api_manager)
  File "/var/lib/juju/agents/unit-k8s-0/charm/./src/charm.py", line 283, in get_node_name
    if self.xcp.name == "aws":
  File "/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/var/lib/juju/agents/unit-k8s-0/charm/venv/charms/interface_external_cloud_provider.py", line 76, in name
    vendor = self._vendor
[...]

PR #259, opened 2025-01-22
Reduce requirements for the model/bundle for tests specifically on openstack

addyess:
### Overview
Reduces the burden/quota necessary to test on openstack down to 2 instances

### Details
* updates the external-cloud-provider (pulls in dmidecode changes for #204)
* ensures fan-networking is disabled when deploying on openstack via integration tests
* shrink the openstack bundle
* references the edge channel openstack integrations
* tests support deploying with trust when necessary


charms/worker/k8s/requirements.txt
```diff
@@ -1,5 +1,5 @@
 charm-lib-contextual-status @ git+https://github.com/charmed-kubernetes/charm-lib-contextual-status@255dd4a23defc16dcdac832306e5f460a0f1200c
-charm-lib-interface-external-cloud-provider @ git+https://github.com/charmed-kubernetes/charm-lib-interface-external-cloud-provider@e1c5fc69e98100a7d43c0ad5a7969bba1ecbcd40
+charm-lib-interface-external-cloud-provider @ git+https://github.com/charmed-kubernetes/charm-lib-interface-external-cloud-provider@e9cadd749ff8e2a6c81fadb0268a8b10e26876be
```
addyess, line 2: support for #204

tests/integration/conftest.py
```diff
@@ -244,11 +244,13 @@ def valid_namespace_name(s: str) -> str:
 
 
 @pytest.fixture()
-@pytest.mark.usefixtures("kubernetes_cluster")
-async def api_client(ops_test: OpsTest, module_name: str):
+async def api_client(
+    kubernetes_cluster, ops_test: OpsTest, module_name: str  # pylint: disable=unused-argument
```
HomayoonAlimohammadi, line 248: Dumb question: why did we do this?

tests/integration/data/test-bundle-openstack.yaml
HomayoonAlimohammadi, line 1: Not a comment for this PR, but in general I think we should move towards using the single source of truth (either bundle or overlay) for these deployments (e.g. `using [this bundle](https://github.com/canonical/k8s-bundles/blob/main/main/bundle.yaml) and applying [this overlay](https://github.com/canonical/k8s-bundles/blob/main/overlays/openstack.yaml)).

tests/integration/data/test-bundle-openstack.yaml
```diff
@@ -8,27 +8,21 @@ series: jammy
...
     expose: true
-    options:
-      bootstrap-node-taints: "node-role.kubernetes.io/control-plane=:NoSchedule"
-  k8s-worker:
-    charm: k8s-worker
-    num_units: 2
-    constraints: cores=2 mem=8G root-disk=16G
   openstack-integrator:
     charm: openstack-integrator
+    channel: edge
```
HomayoonAlimohammadi, line 16: Is this for testing purposes or are we gonna use edge charms for our integration tests?

tests/integration/data/test-bundle-openstack.yaml
```diff
@@ -8,27 +8,21 @@ series: jammy
 applications:
   k8s:
     charm: k8s
-    num_units: 3
+    num_units: 1
     constraints: cores=2 mem=8G root-disk=16G
     expose: true
-    options:
-      bootstrap-node-taints: "node-role.kubernetes.io/control-plane=:NoSchedule"
-  k8s-worker:
```
HomayoonAlimohammadi, line 16: Is there anyway that we can keep deploying a worker charm for tests? I'm a bit scared of removing this coverage. Maybe we can reduce the `num_units` if we have quota issues or...

tests/integration/helpers.py
```diff
@@ -423,6 +424,19 @@ def applications(self) -> Mapping[str, dict]:
...
+    @property
+    def needs_trust(self) -> bool:
+        """Check if the bundle needs to be trusted.
+
+        Returns:
+            bool: True if the bundle needs to be trusted
+        """
+        return any(
+            app.get("trust", False)
+            for app in self.content["applications"].values()
```
HomayoonAlimohammadi, line 436: nit: Maybe an overly defensive programmer here:
```suggestion
            for app in self.content.get("applications", {}).values()
```

HomayoonAlimohammadi (commented):
Great work, thanks a lot @addyess! LGTM overall, left some questions and suggestions.

tests/integration/conftest.py
addyess: I changed this because what I expected from pytest was actually forbidden.  See the warning at the end of this section of the [pytest docs][docs].  Effectively, to use a fixture in another fixture its name MUST be in the arguments list -- it's not enough to mark it like this and in the future this will be an error. 

[docs]:  https://docs.pytest.org/en/stable/how-to/fixtures.html#use-fixtures-in-classes-and-modules-with-usefixtures

tests/integration/data/test-bundle-openstack.yaml
addyess: Thanks for the comment.  I think in the VERY near future we will be deploying using our terraform modules rather than bundles.  All this will be moot

tests/integration/data/test-bundle-openstack.yaml
addyess: well, it's two-fold.  
1) we need the edge versions of these charms because the stable versions don't support Canonical K8s until the 1.32/stable release of charmed kubernetes
2) by sticking with edge we will more quickly find issues in the future.  By testing the bleeding edge of what we produce, we'll find bugs more quikcly

tests/integration/data/test-bundle-openstack.yaml
addyess: the openstack tests specifically don't need a worker for anything.  There are other tests in the PRs that require workers and are still tested. Yes, i did this to shrink the quota requirements on openstack.  If we add specific tests for workers -- i'd be ok to add a worker back in.

tests/integration/helpers.py
addyess: Alright sure.  Check out the new impl

tests/integration/test_openstack.py
```diff
@@ -17,7 +17,7 @@
 STORAGE_CLASS_NAME = "csi-cinder-default"
 
 pytestmark = [
-    pytest.mark.bundle(file="test-bundle-openstack.yaml", apps_local=["k8s", "k8s-worker"]),
+    pytest.mark.bundle(file="test-bundle-openstack.yaml", apps_local=["k8s"]),
```
bschimke95, line 20: why don't we deploy a worker on openstack?

bschimke95 (approved):
LGTM, only one question

tests/integration/test_openstack.py
addyess: the bundle for testing openstack didn't really need a worker.  if there is one that'd be okay.  It was just a matter of making the bundle smaller and quicker to deploy.  I just wanted to make sure a CP node starts and runs the workloads

PR #264, opened 2025-01-27
Add Nightly CI Job for Terraform-Based Deployment and Testing

bschimke95:
#### Summary  
This PR introduces a nightly CI job that deploys the latest edge Kubernetes charms using Terraform and runs integration tests against the deployed clusters.  

#### Rationale  
Initially, we planned to remove charm bundles entirely for the k8s-operator. However, since the Terraform Juju provider does not yet support locally built charms, it’s currently impossible to use Terraform to deploy unpublished charms, such as those in pull requests.  

To address this, we’ve implemented a nightly test that deploys clusters via Terraform and runs the integration test suite. This allows us to test the Terraform modules while maintaining compatibility with our CI workflows.  

#### Changes  
- Added a Python script for deploying Kubernetes clusters using Terraform.  
- Included a default manifest file to deploy a 3-control-plane, 2-worker cluster.  
- Created a `main.tf` file as the root Terraform module.  
- Integrated a nightly GitHub CI job to deploy the cluster and execute the test suite.  

#### Limitations  
This initial implementation:  
- Currently supports only AMD architecture.  
- Runs a basic `all nodes ready` smoke test as the only validation.  

Future updates will address these limitations and expand test coverage.


louiseschmidtgen (commented):
Great work Ben! Two small comments:

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,31 @@
...
+      with:
+        provider: lxd
+        channel: latest/stable
+        juju-channel: 3.6/stable
+    - name: Checking out repo
+      uses: actions/checkout@v4
+    - name: Terraform setup
+      run: |
+        echo "Setting up the cluster with Terraform"
+        python3 ${{ github.workspace }}/tests/integration/terraform_setup.py
```
HomayoonAlimohammadi, line 28: nit: Maybe we can have a `make` command for this one?

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Run integration tests
+      run: |
+        tox -e integration -- -s -k test_nodes_ready --model my-canonical-k8s --no-deploy
```
HomayoonAlimohammadi, line 31: nit: Since (I think) this is temporary and will be replaced by a `make run-e2e-tests` or `tox -e integration` in the future, maybe we can add a TODO comment that explains this.

tests/integration/terraform_setup.py
```diff
@@ -0,0 +1,212 @@
...
+"""
+
+import argparse
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+
+def run_command(command, capture_output=False, fail_on_error=True) -> str | None:
```
HomayoonAlimohammadi, line 31: nit
```suggestion
def run_command(command: str, capture_output: bool = False, fail_on_error: bool = True) -> Optional[str]:
```

tests/integration/terraform_setup.py
```diff
@@ -0,0 +1,212 @@
...
+        )
+        return result.stdout.strip() if capture_output else None
+    except subprocess.CalledProcessError as e:
+        if fail_on_error:
+            print(f"Error running command: {command}\n{e}")
+            sys.exit(1)
+        return None
+
+
+def ensure_terraform(expected_version) -> None:
```
HomayoonAlimohammadi, line 54: nit
```suggestion
def ensure_terraform(expected_version: str) -> None:
```

tests/integration/terraform_setup.py
```diff
@@ -0,0 +1,212 @@
...
+    """Ensure Terraform is installed and matches the expected version.
+
+    Args:
+        expected_version: The expected version of Terraform.
+    """
+    installed_version = run_command(
```
HomayoonAlimohammadi, line 60: I _think_ this will raise an error if Terraform is not installed. If that's the case, maybe we can handle it?

tests/integration/terraform_setup.py
```diff
@@ -0,0 +1,212 @@
...
+            f"does not match the expected version ({expected_version})."
+        )
+        sys.exit(1)
+    else:
+        print(
+            f"Terraform is already installed and matches the expected version: {installed_version}."
+        )
+
+
+def setup_juju_provider_authentication():
```
HomayoonAlimohammadi, line 78: nit
```suggestion
def setup_juju_provider_authentication() -> None:
```

tests/integration/terraform_setup.py
```diff
@@ -0,0 +1,212 @@
...
+        run_command(
+            f"juju show-controller {controller.strip()}"
+            f"| yq '.{controller}.details.\"ca-cert\"' | tr -d '\"' | sed 's/\\\\n/\\n/g'",
+            capture_output=True,
+        )
+        or ""
+    )
+
+
+def ensure_model_exists(model_name, lxd_profile_path):
```
HomayoonAlimohammadi, line 119: nit
```suggestion
def ensure_model_exists(model_name: str, lxd_profile_path: Union[Path, str]) -> None:
```

tests/integration/terraform_setup.py
```diff
@@ -0,0 +1,212 @@
...
+        print("Current Juju controller is not LXD/localhost. Skipping 'k8s.profile' application.")
+
+
+def main():
+    """Main entrypoint."""
+    script_dir = Path(os.path.dirname(os.path.abspath(__file__)))
+
+    parser = argparse.ArgumentParser(description="Terraform and Juju setup script.")
+    parser.add_argument(
+        "--terraform-version", default="latest/stable", help="Expected Terraform version."
```
HomayoonAlimohammadi, line 155: Are these arguments consistent with the module docs at the top?

tests/integration/data/main.tf
```diff
@@ -0,0 +1,32 @@
...
+}
+
+variable "model_name" {
+  description = "Name of the model to deploy to"
+  type        = string
+  default     = "my-canonical-k8s"
+}
+
+module "k8s" {
+  source        = "git::https://github.com/asbalderson/k8s-bundles//terraform?ref=terraform-bundle-basic"
```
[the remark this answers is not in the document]
bschimke95: Fair point, I asked Alex to create a PR to the upstream repo

HomayoonAlimohammadi (commented):
Great work, thanks a lot @bschimke95! Left some minor comments and suggestions.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  integration-tests-terraform:
+    runs-on: self-hosted-linux-amd64-noble-large
+    steps:
+    - name: Setup operator environment
+      uses: charmed-kubernetes/actions-operator@main
+      with:
+        provider: lxd
+        channel: latest/stable
+        juju-channel: 3.6/stable
```
mateoflorido, line 22: We can use the latest `3` version by tracking `3/stable`. Wdyt?
```suggestion
        juju-channel: 3/stable
```

tests/integration/terraform_setup.py
bschimke95: No they are not, I'll remove the args docs above as they eventually run out of sync anyway.

tests/integration/terraform_setup.py
```diff
@@ -0,0 +1,212 @@
...
+            "| yq -r '. | join(\",\")'",
+            capture_output=True,
+        )
+        or ""
+    )
+    os.environ["JUJU_USERNAME"] = (
+        run_command(
+            f"cat ~/.local/share/juju/accounts.yaml"
+            f"| yq .controllers.{controller}.user | tr -d '\"'",
+            capture_output=True,
```
mateoflorido, line 97: Small suggestion regarding `yq`. It should enable the `-r` flag by default, so you won't need the `tr` part.
```suggestion
            f"yq .controllers.{controller}.user < ~/.local/share/juju/accounts.yaml",
            capture_output=True,
```

tests/integration/terraform_setup.py
```diff
@@ -0,0 +1,212 @@
...
+        )
+        or ""
+    )
+    os.environ["JUJU_PASSWORD"] = (
+        run_command(
+            f"cat ~/.local/share/juju/accounts.yaml"
+            f"| yq .controllers.{controller}.password | tr -d '\"'",
+            capture_output=True,
```
mateoflorido, line 105:
```suggestion
            f"yq .controllers.{controller}.password < ~/.local/share/juju/accounts.yaml",
            capture_output=True,
```

mateoflorido (commented):
Small nits around `yq`

.github/workflows/nightly.yaml
bschimke95: Let's add that if we find the need. Right now, the k8s-operator repo does not have any make file setup so I would not do that right now.

tests/integration/terraform_setup.py
bschimke95: Looks like this works just fine if Terraform is not installed:
https://github.com/canonical/k8s-operator/actions/runs/13012467795/job/36293273177#step:5:9

.github/workflows/nightly.yaml
addyess: it has `tox`

tests/integration/terraform_setup.py
addyess: @bschimke95 I love the shell injection available here. There's alot of fun danger with running `subprocess.run(...)` with `shell=True`.  I think we've gotta be more defensive here since this repo has some secrets which could be leaked.

```python
expected_version="latest/edge --classic; sudo snap install app-which-grabs-secrets"
```

addyess: @HomayoonAlimohammadi is kinda right:

```python
>>> subprocess.run("snap list thingy")
  File "<stdin>", line 1, in <module>
  File "/home/addyess/.pyenv/versions/3.8.18/lib/python3.8/subprocess.py", line 493, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/home/addyess/.pyenv/versions/3.8.18/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/home/addyess/.pyenv/versions/3.8.18/lib/python3.8/subprocess.py", line 1720, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'snap list thingy'
>>> subprocess.run("snap list thingy", shell=True)
error: no matching snaps installed
CompletedProcess(args='snap list thingy', returncode=1)
```

Notice how when you run with `shell=True` -- you don't get an error thrown?

tests/integration/terraform_setup.py
addyess: Can we use `yaml`?

```python
accounts = yaml.safe_load(Path("~/.local/share/juju/accounts.yaml").open())
os.envrion["JUJU_USERNAME"] = accounts["controllers"][controller]["user"].strip()
```

tests/integration/terraform_setup.py
```diff
@@ -0,0 +1,192 @@
...
+        lxd_profile_path: The path to the LXD profile file.
+    """
+    # Grep returns 1 if no match is found, so we use `fail_on_error=False` to suppress the error
+    if run_command(
+        f"juju models | grep -q {model_name}", capture_output=True, fail_on_error=False
+    ):
+        print(f"Juju model '{model_name}' already exists.")
+    else:
+        print(f"Juju model '{model_name}' does not exist. Creating it...")
+        run_command(f"juju add-model {model_name} localhost")
```
addyess, line 124: Maybe reference the controller here rather than the cloud `localhost`

```suggestion
        run_command(f"juju add-model -c {controller} {model_name}")
```

addyess (changes requested):
I love the flow, this is going to be nice.  My only beef is the use of `subprocess.run(shell=True)`.  Not everything has to be a shell command.  It doesn't give you the control you are looking for.  For instance, you'll never get a `CalledProcessError` thrown in run_command when you're using `shell=True`.  Something has to change there.

tests/integration/terraform_setup.py
mateoflorido: I agree with @addyess. It would be better to just offload everything to Python constructs.

.github/workflows/nightly.yaml
bschimke95: Fair point, I created a new tox environment.

tests/integration/terraform_setup.py
bschimke95: I ended up using the Juju Python clientlib

addyess (approved):
LGTM 2

PR #265, opened 2025-01-27
Support Multiple Builds-on

mateoflorido:
## Overview
Upgrade the charmcraft.yaml to use `charmcraft: 3.x`, including platform configurations to target multiple bases and architectures.
## Rationale
To accomodate new binary dependencies like `cryptography` and `pydantic-core`, we are switching over `charmcraft 3.x`. This change allows us to build charms for each architecture/release for both the `k8s` and `k8s-worker` charms.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -51,7 +51,7 @@ jobs:
       charmcraft-channel: ${{ needs.charmcraft-channel.outputs.channel }}
       extra-arguments: >-
         ${{needs.extra-args.outputs.args}} -k test_${{ matrix.suite }}
-        ${{ matrix.arch.id == 'arm64' && ' --lxd-containers --series=focal' || '' }}
+        ${{ matrix.arch.id == 'arm64' && ' --lxd-containers --series=jammy' || '' }}
       juju-channel: 3/stable
       load-test-enabled: false
       provider: lxd
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -17,6 +17,7 @@ from typing import Any, Dict, List, Mapping, Optional, Set, Tuple, Union
 import juju.application
 import juju.model
 import juju.unit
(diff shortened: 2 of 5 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
       charmcraft-channel: ${{ needs.charmcraft-channel.outputs.channel }}
       extra-arguments: >-
         ${{needs.extra-args.outputs.args}} -k test_${{ matrix.suite }}
-        ${{ matrix.arch.id == 'arm64' && ' --lxd-containers --series=focal' || '' }}
+        ${{ matrix.arch.id == 'arm64' && ' --lxd-containers --series=jammy' || '' }}
```
addyess, line 54: Remind me again what we're doing here?  Why did this need to switch to jammy?  Is this so we run everything on jammy when we test?

tests/integration/helpers.py
```diff
@@ -433,6 +436,21 @@ def needs_trust(self) -> bool:
...
+        """
+        if not self.series:
+            return "22.04"
+
+        return {
+            "focal": "20.04",
+            "jammy": "22.04",
+            "noble": "24.04",
+        }.get(self.series, "22.04")
+
```
addyess, line 453: while i love this... checkout `juju.utils.get_series_version(...)` and see if it doesn't meet your needs

addyess (approved):
tiny nit.  Love love love the effort here.  There's alot more showing up on the page here than the work put into this PR

.github/workflows/integration_test.yaml
mateoflorido: Yes, I changed this to ensure that we deploy all the tests on jammy.

tests/integration/helpers.py
mateoflorido: I didn't know this was a thing. Updated!

PR #268, opened 2025-01-29
Expose k8s application config via terraform

addyess:
### Overview

The k8s api endpoints are exposed via firewall rules on some substrates through `k8s expose`.  This makes the default to expose the endpoint everywhere by default. 

### Details
The expose block can be configured according to the [juju provider TF docs](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/application#nestedblock--expose)


bschimke95 (approved):
LGTM

PR #274, opened 2025-01-31
e2e tests: Add proxy to model for openstack tests

HomayoonAlimohammadi:
### Overview
This PR adds the necessary model config for the openstack tests to run.


addyess:
you should take a look at just running the tox with [`--apply-proxy`](https://github.com/canonical/k8s-operator/blob/main/tests/integration/conftest.py#L46)

addyess (changes requested):
This seems like a duplicate of an existing feature

[--apply-proxy](https://github.com/canonical/k8s-operator/blob/main/tests/integration/conftest.py#L46)

HomayoonAlimohammadi:
Thanks @addyess, didn't notice this one.

PR #275, opened 2025-01-31
Lint with markdownlint

eaudetcobello:
Lint markdown according to markdownlint rules


addyess (approved):
LGTM.  thanks @eaudetcobello

PR #285, opened 2025-02-04
Streamline storage tests by using k8s python client

addyess:
### Overview
* Replace copying manifest files into k8s cp container to run, with using the kubernetes client api

### Details
* Hopefully improves tests failures occurring with ceph tests

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -103,19 +103,6 @@ def pytest_collection_modifyitems(config, items):
                 item.add_marker(skip_cos)
 
 
-@pytest.fixture(scope="module")
-def module_name(request) -> str:
-    """Get the module name of the test.
-
-    Args:
-        request: Pytest request object.
-
-    Returns:
-        str: The test module name.
-    """
-    return request.module.__name__
-
-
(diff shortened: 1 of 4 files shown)


tests/integration/conftest.py
```diff
@@ -243,29 +230,30 @@ def valid_namespace_name(s: str) -> str:
     return sanitized[-63:]
 
 
-@pytest.fixture()
+@pytest_asyncio.fixture(scope="module")
```
addyess, line 233: without this, pytest doesn't know this is an async fixture.

tests/integration/conftest.py
```diff
@@ -103,19 +103,6 @@ def pytest_collection_modifyitems(config, items):
...
-
-    Args:
-        request: Pytest request object.
-
-    Returns:
-        str: The test module name.
-    """
-    return request.module.__name__
-
-
```
addyess, line 118: This is just silly... why did i put this in here?

eaudetcobello: copilot and <tab><tab><tab>'ll do that

PR #308, opened 2025-02-06
Build per arch, test per module

addyess:
Shrink the number of builds performed on the charms to enable testing from 6 -> 2

* Rather than testing each module with a different set of builds, we can build once per architecture and test with multiple modules
* format and lint with ruff


HomayoonAlimohammadi (approved):
Great work, thanks a lot @addyess! LGTM

mateoflorido (approved):

PR #309, opened 2025-02-06
feat: Use `uv` as the build system

mateoflorido:
## Overview
Use `uv` as the build system for the charms. 
## Changes
This pull request use `uv` as the build system for the `k8s-operator` charms. Creating a `uv.lock` lockfile to track the charms' dependencies.


charms/worker/charmcraft.yaml
```diff
@@ -132,21 +132,27 @@ resources:
...
+      - libffi-dev
+      - libssl-dev
+      - pkg-config
+    build-snaps:
+      - astral-uv
+      - rustup
+    override-build: |
+      rustup default stable
+      for item in pyproject.toml uv.lock lib templates src; do
+        cp -r "$CRAFT_PART_BUILD/k8s/$item" "$CRAFT_PART_BUILD"
```
addyess, line 148: why `copy` and not `move`?

mateoflorido: I've noticed some failures when we use `move`. This is likely because the build directory gets messed up, making it impossible to repack without cleaning the environment (i.e., we already have moved the files). However, this isn't a major problem as the prime step would only keep the `src` and `lib` folders (we add `templates` in the next step).

charms/worker/k8s/uv.lock
```diff
@@ -0,0 +1,619 @@
+version = 1
```
addyess, line 1: Is this auto-generated or is it supposed to be checked in?
I guess checked in is the right thing. I dunno

mateoflorido: Yes, this must be checked in. This would serve as the source of truth for the pinned dependencies. Also, we can receive autoupdates from dependabot.

addyess:
@mateoflorido will you update the CONTRIBUTING guide so i learn how to update the dependencies?  I can't just change the pyproject.toml, i must somehow regenerate this lock file correct?

addyess:
am i right we need for the unit tests to work?

```diff
--- a/charms/worker/k8s/tox.ini
+++ b/charms/worker/k8s/tox.ini
@@ -28,7 +28,7 @@ description = Run unit tests
 deps =
     coverage[toml]
     pytest
-    -r{toxinidir}/requirements.txt
+    .
 commands =
     coverage run --source={[vars]src_path},{[vars]lib_path} \
         -m pytest --ignore={[vars]tst_path}integration -vv \
```

CONTRIBUTING.md
addyess, line 1: I'm happy with the changes here.  style checker warnings or not

charms/worker/k8s/pyproject.toml
```diff
@@ -0,0 +1,33 @@
+[project]
+name = "k8s"
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.8"
```
addyess, line 6: ❤️

addyess (approved):
LGTM

charms/worker/k8s/pyproject.toml
addyess: i wonder if this is no longer needed
https://github.com/canonical/k8s-operator/pull/271

Issue #238, opened 2025-01-09
Containerd Base path moved in classic snap


addyess:
### Bug Description

There was a time that the containerd base path was `/var/snap/k8s/common/etc/containerd/` in all the flavors of the snap, but somewhere around https://github.com/canonical/k8s-snap/pull/817, the path moved to `/etc/containerd` on the classic snap.

The charm didn't change (and had no way to know the containerd base path moved) and therefore now it is broke writing to the wrong directory for custom containerd registries

### To Reproduce

Write a custom registry -- it doesn't do anything
```sh
juju config k8s containerd-custom-registries='[{"url": "http://10.131.231.165:5000", "host": "docker.io"}]
```

### Environment

run it anywhere, its broken all over

### Relevant log output

```shell
N/A
```

addyess:
Resolved with #239

PR #310, opened 2025-02-06
containerd base dir is in the oddest nested spot for reasons...

addyess:
Again address issue #238 

* Honestly, this is the last time it needs to change. 

* We'll need to backport to 1.32


asbalderson:
this makes sense to me +1

eaudetcobello (approved):

PR #313, opened 2025-02-07
Improve external load balancer endpoint testing

HomayoonAlimohammadi:
### Overview
This PR improves testing around external load balancer endpoint.

The actual integration with an external load balancer is also tested in the current Openstack e2e tests. That's because we've added the relation to the test bundle and in the e2e tests we retrieve the kubeconfig by running the `get-kubeconfig` action. A test for that action is added to make sure it returns the external load balancer endpoint when it's available. So the e2e tests intrinsically test the reachability of the cluster via the external lb endpoint.

### DIFF
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -147,6 +147,7 @@ class K8sCharm(ops.CharmBase):
         lead_control_plane: true if this is a control-plane unit and its the leader
         is_upgrade_granted: true if the upgrade has been granted
         datastore: the datastore used for Kubernetes
+        external_load_balancer_address: the external load balancer address, if available
     """
 
     _stored = ops.StoredState()
@@ -187,7 +188,6 @@ class K8sCharm(ops.CharmBase):
         )
         self._upgrade_snap = False
         self._stored.set_default(is_dying=False, cluster_name=str(), upgrade_granted=False)
-        self._external_load_balancer_address = ""
 
         self.cos_agent = COSAgentProvider(
             self,
@@ -207,6 +207,32 @@ class K8sCharm(ops.CharmBase):
             self.kube_control = KubeControlProvides(self, endpoint="kube-control")
             self.framework.observe(self.on.get_kubeconfig_action, self._get_external_kubeconfig)
             self.external_load_balancer = LBProvider(self, EXTERNAL_LOAD_BALANCER_RELATION)
+
+    @property
+    def external_load_balancer_address(self) -> str:
+        """Return the external load balancer address.
+
+        Raises:
+            LookupError: If the loadbalancer response has errors.
+        """
+        if not self.is_control_plane:
+            log.warning("External load balancer is only configured for control-plane units.")
+            return ""
+
+        if not self.external_load_balancer.is_available:
+            log.warning(
+                "External load balancer relation is not available but the address was requested."
+            )
+            return ""
+
+        resp = self.external_load_balancer.get_response(EXTERNAL_LOAD_BALANCER_RESPONSE_NAME)
+        if not resp:
+            log.error("No response from external load balancer")
+            return ""
+        if resp.error:
+            raise LookupError(f"External load balancer error: {resp.error}")
+
+        return resp.address
 
     def _k8s_info(self, event: ops.EventBase):
         """Send cluster information on the kubernetes-info relation.
@@ -387,7 +413,11 @@ class K8sCharm(ops.CharmBase):
         self.api_manager.check_k8sd_ready()
 
     def _get_extra_sans(self):
-        """Retrieve the certificate extra SANs."""
+        """Retrieve the certificate extra SANs.
+
+        Raises:
+            ReconcilerError: If it fails to get the external load balancer address.
+        """
         # Get the extra SANs from the configuration
         extra_sans_str = str(self.config.get("kube-apiserver-extra-sans") or "")
         extra_sans = set(extra_sans_str.strip().split())
@@ -401,9 +431,12 @@ class K8sCharm(ops.CharmBase):
             extra_sans |= {str(addr) for addr in addresses}
 
         # Add the external load balancer address
-        if self._external_load_balancer_address:
-            log.info("Adding external load balancer address to extra SANs")
-            extra_sans.add(self._external_load_balancer_address)
+        try:
+            if lb_addr := self.external_load_balancer_address:
+                log.info("Adding external load balancer address to extra SANs")
+                extra_sans.add(lb_addr)
+        except LookupError as e:
+            raise ReconcilerError(f"Failed to get external load balancer address: {e}")
 
         return sorted(extra_sans)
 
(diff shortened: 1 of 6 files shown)


charms/worker/k8s/src/charm.py
```diff
@@ -206,6 +206,34 @@ def __init__(self, *args):
...
+
+        resp = self.external_load_balancer.get_response(EXTERNAL_LOAD_BALANCER_RESPONSE_NAME)
+        if not resp:
+            log.error("No response from external load balancer")
+            return ""
+        if resp.error:
+            log.error("External load balancer error: %s", resp.error)
+            return ""
+
+        self._external_load_balancer_address = resp.address
```
bschimke95, line 235: What if the external loadbalancer relation changes (and with that, the address). 
It would still return the old address. Can this happen?

bschimke95 (commented):
Generally LGTM, only a small question

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: I might be wrong, but I think then the relation data changes, the charm executes again and this attribute gets emptied and re-evaluated.
We only keep this tiny caching mechanism for a single charm execution and AFAIK we can't experience these kinds of relation data changes in the middle and execution.
WDYT?

HomayoonAlimohammadi:
At the moment we don't have any automated way of running the Openstack integration tests in the CI, I ran them manually and they all pass. The external load balancer gets created and the cluster is accessible using the load balancer endpoint.

charms/worker/k8s/src/charm.py
bschimke95: Right, thanks for the context.

bschimke95 (approved):

tests/integration/data/test-bundle-openstack.yaml
```diff
@@ -27,4 +27,5 @@ relations:
   - [cinder-csi:kube-control,                            k8s:kube-control]
   - [openstack-cloud-controller:external-cloud-provider, k8s:external-cloud-provider]
   - [openstack-cloud-controller:openstack,               openstack-integrator:clients]
-  - [cinder-csi:openstack,                               openstack-integrator:clients]
\ No newline at end of file
+  - [cinder-csi:openstack,                               openstack-integrator:clients]
+  - [k8s:external-load-balancer,                         openstack-integrator:lb-consumers]
```
addyess, line 31: does the `test_openstack.py` have a new test to contact the api server using the load-balanced address?  It might be as trivial as confirming the endpoint we're using is not the IP address of any of the k8s units.

addyess (changes requested):
I'd love to see the charm block if we get an error from the openstack-integrator

charms/worker/k8s/src/charm.py
```diff
@@ -206,6 +206,34 @@ def __init__(self, *args):
...
+            return ""
+
+        if not self.external_load_balancer.is_available:
+            log.warning(
+                "External load balancer relation is not available but the address was requested."
+            )
+            return ""
+
+        if self._external_load_balancer_address:
+            return self._external_load_balancer_address
```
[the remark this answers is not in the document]
HomayoonAlimohammadi: That's right, thanks a lot for the explanation. So I guess it's best to use `@cached_property` instead of introducing this additional attribute.
Maybe it's not even that useful, an unnecessary complexity. Since we probably only fetch the lb endpoint address a few times in each run.

tests/integration/data/test-bundle-openstack.yaml
HomayoonAlimohammadi: I think in [this part](https://github.com/canonical/k8s-operator/blob/main/tests/integration/helpers.py#L221-L244) we're getting the kubeconfig through `get-kubeconfig` action and this action [uses the "most" public address](https://github.com/canonical/k8s-operator/blob/main/charms/worker/k8s/src/charm.py#L1151). The "most" public address [returns the external lb endpoint](https://github.com/canonical/k8s-operator/blob/main/charms/worker/k8s/src/charm.py#L1181-L1183) when possible.
Doesn't this make all the Openstack integration tests to contact the apiserver through the external lb endpoint?

HomayoonAlimohammadi:
Running `tox run -e format` changed many many things not introduced in my PR...

HomayoonAlimohammadi:
Addressed the formatting changes in https://github.com/canonical/k8s-operator/pull/321

charms/worker/k8s/src/charm.py
```diff
@@ -1177,10 +1215,16 @@ def _get_public_address(self) -> str:
...
         """
-        if self._external_load_balancer_address:
-            log.info("Using external load balancer address as the public address")
-            return self._external_load_balancer_address
+        try:
+            if lb_addr := self.external_load_balancer_address:
+                log.info("Using external load balancer address as the public address")
+                return lb_addr
+        except LookupError as e:
+            raise LookupError("Failed to get external load balancer address: %s", e)
```
addyess, line 1227: nit: we're catching the original exception, only to drop the stacktrace here.  Lets preserve the original stacktrace


```suggestion
        except LookupError as e:
            raise LookupError("Failed to get external load balancer address") from e
```

addyess (commented):
Small nit... then its LGTM. 

I would love @mateoflorido to evaluate with respect to TLS integration and the extra-sans adjustments

addyess:
> @addyess Thanks for the review! Addressed the comments and added two integration tests to both check the external load balancer and ensure certificates change correctly.

Looks good -- got some conflicts to resolve too

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: I don't fully understand what "dropping the stacktrace" means here. In this example:
```python
try:
    raise LookupError("initial")
except LookupError as e:
    raise LookupError(f"wrapper: {e}")
```
We'll get this output:
```
  File "<stdin>", line 2, in <module>
LookupError: initial

During handling of the above exception, another exception occurred:

  File "<stdin>", line 4, in <module>
LookupError: wrapper: initial
```
So I _think_ we do have the stacktrace here, don't we? (During handling of the above exception...)

addyess: ok, So python doesn't swallow the first exception, but the right way to [chain][0] python exceptions like this is with the `from`.  Here's "why".  In one case it looks like an exception occurred while handling(catching) the first exception.  In the second case, it's clear that we're wrapping or chaining an exception.  Note the subtle difference in the stack trace messages:

1) `During handling of the above exception, another exception occurred`
2) `The above exception was the direct cause of the following exception`

By using from you provide more context to future catching up the chain

[0]: https://docs.python.org/3/tutorial/errors.html#exception-chaining

HomayoonAlimohammadi: Smart Adam, smart as always. I didn't notice the subtle difference. Nice!

HomayoonAlimohammadi: I changed it to
```python
except LookupError as e:
    raise LookupError(f"parent error: {e}") from e
```
I think this way we can have the best of both worlds :D
WDYT?

addyess (approved):
Awesome work @HomayoonAlimohammadi . You're tenacious!

PR #317, opened 2025-02-08
Catch httpx.ConnectError as a cluster-inspect error

addyess:
fix: catch connection errors in update-status

Encountered this error in a [pr run](https://github.com/canonical/k8s-operator/actions/runs/13218324114/job/36900410921?pr=316) and [here](https://github.com/canonical/k8s-operator/actions/runs/13218325363/job/36900429445?pr=301).  It seems like a bit of a race condition, but the charm erred so let's look out for these situations.


HomayoonAlimohammadi (approved):
Thanks a lot @addyess! LGTM

PR #319, opened 2025-02-10
feat(config): Prevent changing bootstrap charm config

HomayoonAlimohammadi:
### Overview
This PR adds a way to prevent the `bootstrap-` charm configs from getting changed after the cluster is bootstrapped.

### Rationale
Some of the charm config options start with `bootstrap-` which means that these options should only be specified upon deploying the charm (before the underlying Kubernetes cluster is bootstrapped) and changing them afterwards might result in unexpected behavior. We've discussed that we like to have a mechanism to prevent the user from changing these options.


HomayoonAlimohammadi:
Converting this draft as we need to implement the k8sd call for bootstrap config first.

charms/worker/k8s/templates/snap_installation.yaml
```diff
@@ -5,9 +5,9 @@ amd64:
   - name: k8s
     install-type: store
     classic: true
-    channel: latest/edge
+    channel: latest/edge/hue-cluster-config
```
HomayoonAlimohammadi, line 8: This is for testing purposes, needs to be removed before merge.

charms/worker/k8s/charmcraft.yaml
```diff
@@ -86,11 +86,11 @@ config:
         Example:
           e.g.: key1=value1 key2=value2
     bootstrap-datastore:
-      default: dqlite
+      default: k8s-dqlite
       type: string
       description: |
         The datastore to use in Canonical Kubernetes. This cannot be changed
-        after deployment. Allowed values are "dqlite" and "etcd". If "etcd" is
+        after deployment. Allowed values are "k8s-dqlite" and "etcd". If "etcd" is
```
addyess, line 93: I don't think we can just change this.  charms deployed from 1.32 will deploy with `dqlite` as their default.  Upgrading to this revision will cause an immediate failure.

  The default config from 1.32 was dqlite, so that's what we are "stuck with"

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -87,15 +87,31 @@ class InvalidResponseError(K8sdAPIManagerError):
         code (int): HTTP Status code
     """
 
-    def __init__(self, code: int, msg: str) -> None:
```
addyess, line 90: there's a `LIBPATCH` version at the head of this file that if it gets changed, we must also bump

charms/worker/k8s/src/charm.py
```diff
@@ -1310,14 +1323,124 @@ def _ensure_cert_sans(self):
...
+            )
+        except InvalidResponseError as e:
+            if e.code == HTTP_503_SERVICE_UNAVAILABLE:
+                log.info("k8sd is not ready, skipping bootstrap config check")
+                return
+            raise
+
+        # NOTE(Hue): The way we're saving certificates has nothing to do with the snap
+        # so we can't rely on the snap to tell us if the certificates have changed.
+        self._stored.set_default(certificates=self.config.get(CONFIG_BOOTSTRAP_CERTIFICATES))
```
addyess, line 1359: so we're back to using stored-state?

addyess (commented):
Well, i have a few questions.  
I'd also like @mateoflorido to weigh in here too

charms/worker/k8s/charmcraft.yaml
mateoflorido: I agree with Adam. Also, I think this reflect better that we are using `dqlite` as the datastore. The translation mechanism is something that we can hide internally.

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -838,6 +858,34 @@ class RefreshCertificatesRunResponse(BaseRequestModel):
...
+        status (UserFacingClusterConfig): The configuration of the cluster.
+        datastore (UserFacingDatastoreConfig): The configuration of the datastore.
+        node_taints (List[str]): List of node taints.
+    """
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    status: UserFacingClusterConfig
+    datastore: Optional[UserFacingDatastoreConfig] = Field(default=None)
+    node_taints: Optional[List[str]] = Field(default=None, alias="nodeTaints")
```
mateoflorido, line 874: This alias looks odd... Are we going to change the naming schema?

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
```diff
@@ -942,6 +990,8 @@ def create_connection(self):
 class K8sdAPIManager:
     """Manager for K8sd API interactions."""
 
+    CLUSTER_CONFIG_ENDPOINT: str = "/1.0/k8sd/cluster/config"
```
mateoflorido, line 993: Well, we can keep it inline or create another PR to refactor all the other endpoints. It would be a good opportunity to improve the library, but I think it would make the PR harder to review.

charms/worker/k8s/src/charm.py
```diff
@@ -1310,14 +1323,124 @@ def _ensure_cert_sans(self):
...
+                )
+            )
+
+        if changed:
+            for c in changed:
+                log.error(
+                    "Bootstrap config '%s' should NOT be changed after bootstrap. %s", c.name, c
+                )
+            raise BootstrapConfigChangeError(changed)
+
```
addyess, line 1443: Could i talk you into moving all of this out of charm.py.  Maybe bootstrap.py?

charms/worker/k8s/src/charm.py
mateoflorido: We can infer this from the certificates in the unit:
```python
def check_ca_key():
    return os.path.isFile("/etc/kubernetes/pki/ca.key")
```
This way we can avoid using stored state

charms/worker/k8s/src/charm.py
mateoflorido: Agree, `charm.py` is becoming a behemoth

charms/worker/k8s/charmcraft.yaml
HomayoonAlimohammadi: You're right. I also thought that this might be a breaking change, but here's what I think:
We use the `dqlite` string in the following places:
- https://github.com/canonical/k8s-operator/blob/main/charms/worker/k8s/src/charm.py#L729
- https://github.com/canonical/k8s-operator/blob/main/charms/worker/k8s/src/upgrade.py#L211
- https://github.com/canonical/k8s-operator/blob/main/charms/worker/k8s/src/literals.py#L26
And we're not communicating that string with any other source and it's completely contained in the charm. So we're technically not breaking anything _but_ the user experience. At least this is what I assume. 
Or maybe the charm config is preserved during an upgrade. If so, that's breaking.
I'll change this back to `dqlite` anyways.

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
HomayoonAlimohammadi: I'll remove this and create a card for it.

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: Of course you can :D

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
HomayoonAlimohammadi: As I always forget this. Lemme create a card for adding a CI check for this.

charms/worker/k8s/lib/charms/k8s/v0/k8sd_api_manager.py
HomayoonAlimohammadi: I changed this in `k8s-snap-api`, `k8s-snap` and also here.

charms/worker/k8s/src/charm.py
```diff
@@ -1340,6 +1358,43 @@ def _on_refresh_certs_action(self, event: ops.ActionEvent):
             except (InvalidResponseError, K8sdConnectionError) as e:
                 event.fail(f"Failed to refresh certificates: {e}")
 
+    @on_error(
+        ops.BlockedStatus("Cannot change bootstrap config after bootstrap. Check logs."),
+        BootstrapConfigChangeError,
+    )
+    @on_error(
+        ops.ErrorStatus("Failed to get communicate with k8sd."),
```
addyess, line 1366: I don't think its really meant for the charm to use `ops.ErrorStatus`.  Usually this is reserved for the jujud agent to raise.  @mateoflorido could correct me but i think the charm should only ever raise.  That's because an error condition tells the user they MUST `juju resolve <unit>` in order for it to execute anymore hooks.  Raising an error essentially is the charm giving up working until the user intervenes -- that could be years. 

stick to: maintenance, active, waiting, and blocked.  

In which case -- this COULD be blocked.  OR it could be waiting.  K8sd didn't answer?  My guess is it would on the next hook.  Go with `Waiting` here.

charms/worker/k8s/src/charm.py
```diff
@@ -1340,6 +1358,43 @@ def _on_refresh_certs_action(self, event: ops.ActionEvent):
             except (InvalidResponseError, K8sdConnectionError) as e:
                 event.fail(f"Failed to refresh certificates: {e}")
 
+    @on_error(
+        ops.BlockedStatus("Cannot change bootstrap config after bootstrap. Check logs."),
+        BootstrapConfigChangeError,
+    )
```
addyess, line 1364: perfect... the user configured something and that's not valid to change.  Tell them to fix their configuration

charms/worker/k8s/src/charm.py
```diff
@@ -1340,6 +1358,43 @@ def _on_refresh_certs_action(self, event: ops.ActionEvent):
...
+        else:
+            try:
+                ref_config = BootstrapConfigOptions.build(
+                    node_status=self.api_manager.get_node_status().metadata,
+                )
+            except InvalidResponseError as e:
+                if e.code == http.HTTPStatus.SERVICE_UNAVAILABLE:
+                    log.info("k8sd is not ready, skipping bootstrap config check")
+                    return
+                raise
```
addyess, line 1394: Why do we need to catch the exception? Just for the info level log.  The reconciler will log for us already when the exception is raised. 

Are you saying that if the `build(...)` method really can raise an error and everything is fine and the reconciler should march on?  i notice the `e.code == http.HTTPStatus.SERVICE_UNAVAILABLE`.  What's this situation where k8ds is just kinda not ready and we're ok skipping the bootstrap config check.  I'd rather have the thing fail

```suggestion
        ref_config = BootstrapConfigOptions.build(
            cluster_config=(
                self.api_manager.get_cluster_config().metadata
                if self.is_control_plane
                else None
            ),
            node_status=self.api_manager.get_node_status().metadata,
        )
```

charms/worker/k8s/src/config/bootstrap.py
```diff
@@ -0,0 +1,237 @@
...
+        description=f"Represents {CONFIG_BOOTSTRAP_CERTIFICATES} config option."
+    )
+    datastore: str = Field(description=f"Represents {CONFIG_BOOTSTRAP_DATASTORE} config option.")
+    node_taints: str = Field(
+        description=f"Represents {CONFIG_BOOTSTRAP_NODE_TAINTS} config option."
+    )
+    pod_cidr: str = Field(description=f"Represents {CONFIG_BOOTSTRAP_POD_CIDR} config option.")
+    service_cidr: str = Field(
+        description=f"Represents {CONFIG_BOOTSTRAP_SERVICE_CIDR} config option."
+    )
```
addyess, line 66: do we know when these descriptions ever show up for the user?  Are they just here for our own benefit?

addyess (commented):
Just a few suggestions for slimming things down.

charms/worker/k8s/src/config/bootstrap.py
HomayoonAlimohammadi: I don't think the users are ever going to see this. Lemme remove this to prevent unnecessary confusion.

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: I think it can happen when a cluster is not bootstrapped. Maybe we can play around with the ordering of the methods we call in the `_reconcile`.

addyess: well, can we at least collapse this  `if...try...except...else...try...except` down to `try...except`


```suggestion
        try:
            ref_config = BootstrapConfigOptions.build(
                cluster_config=self.is_control_plane and self.api_manager.get_cluster_config().metadata,
                node_status=self.api_manager.get_node_status().metadata,
            )
        except InvalidResponseError as e:
            if e.code == http.HTTPStatus.SERVICE_UNAVAILABLE:
                log.info("k8sd is not ready, skipping bootstrap config check")
                return
            raise
```

addyess (approved):
Small nit.  Looks better though

charms/worker/k8s/src/charm.py
HomayoonAlimohammadi: Great suggestion. It just needs some small changes tho, because `get_node_status` is restricted for worker nodes.
Changed it to:
```python
        try:
            ref_config = BootstrapConfigOptions.build(
                node_status=self.api_manager.get_node_status().metadata,
                cluster_config=self.api_manager.get_cluster_config().metadata
                if self.is_control_plane
                else None,
            )
        except InvalidResponseError as e:
            if e.code == http.HTTPStatus.SERVICE_UNAVAILABLE:
                log.info("k8sd is not ready, skipping bootstrap config check")
                return
            raise
```

PR #321, opened 2025-02-10
chore: tox run -e format

HomayoonAlimohammadi:
### Overview
Seems like our formatting tool makes new changes since we've switched to `uv`. The changes in this PR are only made by running `tox run -e format`.


addyess (approved):
Yes, We do need this..

PR #328, opened 2025-02-12
test: Add integration tests for Kubernetes docker-registry deployment

eaudetcobello:
Adds a test that deploys the docker-registry charm, pulls an image to the registry, tags it with a unique name, and finally tries to pull the image from the registry.

This validates that the charm sets the custom_containerd_config from charm configuration in the right place on the filesystem.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -38,18 +38,18 @@ jobs:
     strategy:
       matrix:
         arch:
-        # built on azure, test on self-hosted
-        - id: amd64
-          builder-label: ubuntu-22.04
-          tester-arch: AMD64
-          tester-size: xlarge
-          modules: '["test_k8s", "test_etcd", "test_ceph", "test_upgrade", "test_external_certs", "test_registry"]'
-        # built and test on on self-hosted
-        - id: arm64
-          builder-label: ARM64
-          tester-arch: ARM64
-          tester-size: large
-          modules: '["test_k8s", "test_etcd"]'
+          # built on azure, test on self-hosted
+          - id: amd64
+            builder-label: ubuntu-22.04
+            tester-arch: AMD64
+            tester-size: xlarge
+            modules: '["test_k8s", "test_etcd", "test_ceph", "test_upgrade", "test_external_certs", "test_registry"]'
+          # built and test on on self-hosted
+          - id: arm64
+            builder-label: ARM64
+            tester-arch: ARM64
+            tester-size: large
+            modules: '["test_k8s", "test_etcd"]'
     with:
       identifier: ${{ matrix.arch.id }}
       builder-runner-label: ${{ matrix.arch.builder-label }}
@@ -64,7 +64,7 @@ jobs:
       self-hosted-runner: true
       self-hosted-runner-arch: ${{ matrix.arch.tester-arch }}
       self-hosted-runner-label: ${{ matrix.arch.tester-size }}
-      test-timeout: 120
+      test-timeout: 600
       test-tox-env: integration
       trivy-fs-enabled: false
       trivy-image-config: "trivy.yaml"
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -1,4 +1,3 @@
-
 name: Integration tests
 
 on:
@@ -28,9 +27,9 @@ jobs:
     outputs:
       channel: ${{ steps.charmcraft.outputs.channel }}
     steps:
-    - uses: actions/checkout@v4
-    - id: charmcraft
-      run: echo "channel=$(cat .charmcraft-channel)" >> $GITHUB_OUTPUT
+      - uses: actions/checkout@v4
+      - id: charmcraft
+        run: echo "channel=$(cat .charmcraft-channel)" >> $GITHUB_OUTPUT
 
   integration-tests:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@a23cb8af2877316bd87af472b1732318386bd05c
@@ -44,7 +43,7 @@ jobs:
           builder-label: ubuntu-22.04
           tester-arch: AMD64
           tester-size: xlarge
-          modules: '["test_k8s", "test_etcd", "test_ceph", "test_upgrade", "test_external_certs"]'
+          modules: '["test_k8s", "test_etcd", "test_ceph", "test_upgrade", "test_external_certs", "test_registry"]'
         # built and test on on self-hosted
         - id: arm64
           builder-label: ARM64
--- a/tests/integration/test_registry.py
+++ b/tests/integration/test_registry.py
@@ -59,9 +59,9 @@ async def test_custom_registry(kubernetes_cluster: model.Model, api_client):
     )
     test_pod_manifest[0]["metadata"]["name"] = random_pod_name
 
-    test_pod_manifest[0]["spec"]["containers"][0][
-        "image"
-    ] = f"{docker_registry_ip}:5000/busybox:latest"
+    test_pod_manifest[0]["spec"]["containers"][0]["image"] = (
+        f"{docker_registry_ip}:5000/busybox:latest"
+    )
 
     k8s_unit = kubernetes_cluster.applications["k8s"].units[0]
     try:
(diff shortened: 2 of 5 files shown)


addyess:
@eaudetcobello can you add `test_registry` as another tested module for amd64

* look in `.github/workflows/integration_test.yaml` and find around line 47

```
          modules: '["test_k8s", "test_etcd", "test_ceph", "test_upgrade"]'
```

add your module (`test_registry`) in there.

tests/integration/test_registry.py
```diff
@@ -0,0 +1,89 @@
...
+        f"{docker_registry_ip}:5000/busybox:latest"
+    )
+
+    k8s_unit = kubernetes_cluster.applications["k8s"].units[0]
+    try:
+        created.extend(*create_from_yaml(api_client, yaml_objects=test_pod_manifest))
+        await helpers.wait_pod_phase(k8s_unit, random_pod_name, "Running")
+    finally:
+        # Cleanup
+        for resource in reversed(created):
```
mateoflorido, line 85: nit: This only deploys a Pod, right? Is this `reversed` required?
```suggestion
        for resource in created:
```

tests/integration/test_registry.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    test_pod_manifest[0]["metadata"]["name"] = random_pod_name
+
+    test_pod_manifest[0]["spec"]["containers"][0]["image"] = (
+        f"{docker_registry_ip}:5000/busybox:latest"
+    )
+
+    k8s_unit = kubernetes_cluster.applications["k8s"].units[0]
+    try:
+        created.extend(*create_from_yaml(api_client, yaml_objects=test_pod_manifest))
```
mateoflorido, line 81: `create_from_yaml` [returns](https://github.com/kubernetes-client/python/blob/de280fbf1ffcf625131f5449a21cc635cdc1169a/kubernetes/utils/create_from_yaml.py#L164) a list of k8s object. Which might cause a TypeError, because extend expects an iterable (like a list or a tuple). You can pass the `create_from_yaml` call directly, or just assign the value to created, as you don't have multiple YAML definitions in your Pod manifest.
```suggestion
        created = create_from_yaml(api_client, yaml_objects=test_pod_manifest)
```

mateoflorido (approved):
LGTM. small nit regarding a potential `TypeError` in the code.

tests/integration/test_registry.py
```diff
@@ -0,0 +1,89 @@
...
+    k8s_app = kubernetes_cluster.applications["k8s"]
+
+    assert k8s_app
+
+
+@pytest.mark.abort_on_fail
+async def test_custom_registry(kubernetes_cluster: model.Model, api_client):
+    """Test that the charm configures the correct directory and can access a custom registry."""
+    # List of resources created during the test
+    created: List = []
```
bschimke95, line 49: nit: `List[dict]`

tests/integration/test_registry.py
```diff
@@ -0,0 +1,89 @@
...
+    """Retrieve the full path of the specified test data file."""
+    return Path(__file__).parent / "data" / "test_registries" / name
+
+
+@pytest.mark.abort_on_fail
+async def test_ready_nodes(kubernetes_cluster: model.Model):
+    # Check that the units are ready
+    k8s_app = kubernetes_cluster.applications["k8s"]
+
+    assert k8s_app
```
bschimke95, line 42: I don't think this is a sufficient check. See https://github.com/canonical/k8s-operator/blob/433a4fdbdb1f52a701c0c48455bc88a87e04b5e4/tests/integration/test_k8s.py#L46

tests/integration/test_registry.py
```diff
@@ -0,0 +1,89 @@
...
+log = logging.getLogger(__name__)
+
+
+def _get_data_file_path(name) -> Path:
+    """Retrieve the full path of the specified test data file."""
+    return Path(__file__).parent / "data" / "test_registries" / name
+
+
+@pytest.mark.abort_on_fail
+async def test_ready_nodes(kubernetes_cluster: model.Model):
```
bschimke95, line 38: nit: we have a bunch of `test_nodes_ready` tests - let's be consistent with that.

tests/integration/test_registry.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+    test_pod_manifest[0]["metadata"]["name"] = random_pod_name
+
+    test_pod_manifest[0]["spec"]["containers"][0]["image"] = (
+        f"{docker_registry_ip}:5000/busybox:latest"
+    )
+
+    k8s_unit = kubernetes_cluster.applications["k8s"].units[0]
+    try:
+        created = create_from_yaml(api_client, yaml_objects=test_pod_manifest)
```
bschimke95, line 81: `create_from_yaml` is an async function, right?
Should this be `created = await create_from_yaml(...)` then?

bschimke95 (approved):
Mostly LGTM, minor comments - great work!

tests/integration/test_registry.py
eaudetcobello: No, not async

tests/integration/test_registry.py
eaudetcobello: Sure

tests/integration/test_registry.py
eaudetcobello: Done

addyess (changes requested):
Using that kubernetes api is a little bonkers at time...here's a suggestion

.github/workflows/integration_test.yaml
```diff
@@ -44,7 +43,7 @@ jobs:
           builder-label: ubuntu-22.04
           tester-arch: AMD64
           tester-size: xlarge
-          modules: '["test_k8s", "test_etcd", "test_ceph", "test_upgrade", "test_external_certs"]'
+          modules: '["test_k8s", "test_etcd", "test_ceph", "test_upgrade", "test_external_certs", "test_registry"]'
```
bschimke95, line 46: @addyess why do we need to explicitly list each test? That feels like it just waits to be forgotten.

tests/integration/test_registry.py
```diff
@@ -0,0 +1,81 @@
...
+
+    await kubernetes_cluster.applications["k8s"].set_config(custom_registry_config)
+    await kubernetes_cluster.wait_for_idle(status="active")
+
+    action = await docker_registry_unit.run_action("push", image="busybox:latest", pull=True)
+    await action.wait()
+
+    # Create a pod that uses the busybox image from the custom registry
+    # Image: {docker_registry_ip}:5000/busybox:latest
+    test_pod_manifest = list(yaml.safe_load_all(_get_data_file_path("pod.yaml").open()))
```
bschimke95, line 60: let's inline this path. It is only used once.

bschimke95 (commented):
Couple of minor things, otherwise LGTM

.github/workflows/integration_test.yaml
addyess: The operator workflows run a unique gh runner for every item in this list.  We run all the tests on a unique runner for these 5 test modules b/c they deploy 5 different SKUs. If we pushed the `registry` tests into another module, they wouldn't have to be added to this list.  you COULD push this into the `test_k8s` module. Or leave it as is and add this modules here. 

we could *detect* this list i suppose.  Not good for this PR though

tests/integration/test_registry.py
```diff
@@ -0,0 +1,81 @@
...
+pytestmark = [
+    pytest.mark.bundle(file="test_registries/test-bundle-docker-registry.yaml", apps_local=["k8s"])
+]
+
+log = logging.getLogger(__name__)
+
+
+def _get_data_file_path(name) -> Path:
+    """Retrieve the full path of the specified test data file."""
+    return Path(__file__).parent / "data" / "test_registries" / name
```
addyess, line 34: Yep, this can just be  a `CONSTANT`

```suggestion
TEST_DATA_PATH = Path(__file__).parent / "data" / "test_registries" / name
```

tests/integration/test_registry.py
addyess:
```suggestion
    test_pod_manifest = list(yaml.safe_load_all(TEST_DATA_PATH.open()))
```

addyess (approved):
Looking better

PR #332, opened 2025-02-14
Pin `kube-control` dependency

mateoflorido:
## Overview
`kube-control` is now a Python package, let's pin it to the current working version.


addyess (approved):
YES PLEASE

PR #339, opened 2025-02-18
Allow Juju tf provider to float past 0.14.0

addyess:
fix: be less specific on juju version of terraform provider to facilitate deployment with `juju >= 0.14.0`


bschimke95 (approved):
LGTM

PR #345, opened 2025-02-20
Bypass operator-workflows promote

addyess:
fix:  charm promotion job needs to be more aware of all the revisions of the charms to promote

### Details

The operator-workflows are limited in how it promotes charms.  It can only promote one series/arch revision to the target channel rather than ALL of the revisions in a certain channel.

The workflows do allow one to specify which series/arch they wish to promote, but fails to "validate" that series and arch as it expects a charmcraft 2.x style "bases" header in the charmcraft.yaml rather than the charmcraft 3.x style. 

By switching to the `charming-actions/promote-charm` workflow, it does a better job of just publishing whatever charms are in the the origin channel to the destination channel along with the charm revisions.  It's all wee really need here.


HomayoonAlimohammadi (approved):
Thanks for digging into this @addyess, LGTM

PR #352, opened 2025-02-24
fix: change subprocess calls for cryptography

mateoflorido:
## Overview
Migrate from subprocess calls to `cryptography`

### DIFF
--- a/charms/worker/k8s/src/pki.py
+++ b/charms/worker/k8s/src/pki.py
@@ -51,4 +51,4 @@ def get_certificate_sans(cert_path: Union[str, Path]) -> Tuple[List[str], List[s
         log.exception("No SAN extension found in certificate %s", cert_path)
         return [], []
 
-    return list(dns_sans), list(ip_address)
+    return sorted(dns_sans), sorted(ip_address)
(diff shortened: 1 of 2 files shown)


charms/worker/k8s/src/pki.py
```diff
@@ -6,57 +6,49 @@
...
-            ip_str = san[len(ip_prefix):]
-            try:
-                ip = ip_address(ip_str)
-                ip_sans.add(str(ip))
-            except ValueError:
-                log.warning("Invalid IP SAN: %s", ip_str)
-        # fmt: on
-
-    return list(dns_sans), list(ip_sans)
+    return list(dns_sans), list(ip_address)
```
addyess, line 54: How about we sort them.  Is that okay?

```suggestion
    return sorted(dns_sans), sorted(ip_address)
```

addyess (approved):
Yeah, only a tiny nit about sorting the result

charms/worker/k8s/src/pki.py
mateoflorido: Sure!

PR #357, opened 2025-02-25
Update K8s snap revisions in a single PR rather than multiple prs

addyess:
When snap revisions are released to the stable channels, we always get two PRs, one that ends in a conflict. 

let's just update all the snaps in a single PR for each branch


bschimke95 (approved):

bschimke95:
LGTM after conflicts are resolved

PR #373, opened 2025-02-26
Only run renovate on main

addyess:
Ok, let's only renovate on main

any security issues we'll backport when necessary


mateoflorido (approved):

addyess:
> Hmmm. I think something failed in the Charm Analysis. The `Dockerfile lint` step should be unrelated...

Fixed!

PR #395, opened 2025-03-11
fix: refactor `external-certificates` integration

mateoflorido:
Applicable spec: **KU123**

### Overview

Refactor the `tls-certificates` relation to request external facing certificates.

### Rationale

As outlined in the spec, we now only use the `tls-certificates` relation for external facing components, such as `kube-apiserver` and `kubelet`. Internal facing certificates (or client certs) are now managed internally by the `k8s` snap's machinery. 

### Juju Events Changes

None

### Module Changes

- Add formatters to external facing certificates' common names
- Simplified the CSR logic in the `certificates` module.
- Added validators to enforce certificate conditions (no empty or invalid formatters, restricted CN length per [RFC 5280](https://www.rfc-editor.org/rfc/rfc5280))
- Use the certificate chain to generate a valid CA certificate.


charms/worker/k8s/src/certificates.py
```diff
@@ -154,50 +176,31 @@ def _certificates_request_mapping(
...
-            return {k: v for k, v in csr_attributes.items() if k in CONTROL_PLANE_CERTIFICATES}
-        return {k: v for k, v in csr_attributes.items() if k in WORKER_CERTIFICATES}
+        fmt_context = {
+            "node-name": self._charm.get_node_name(),
+            "cluster-name": self._charm.get_cluster_name(),
+            "domain-name": self._charm.config.get("external-certs-domain-name"),
+        }
+        try:
+            formatted_name = formatter.format(**fmt_context)
+            if not formatted_name:
```
HomayoonAlimohammadi, line 198: Nice! I couldn't have caught this one

charms/worker/k8s/src/certificates.py
```diff
@@ -209,8 +212,13 @@ def _get_service_ips(self) -> Set[str]:
             ValueError: If the service CIDR is invalid.
         """
         service_ips = set()
-        service_cidrs = self._charm.config.get("bootstrap-service-cidr", "").split(",")
-        for cidr in service_cidrs:
+        service_cidrs = self._charm.config.get("bootstrap-service-cidr", "")
+        if not isinstance(service_cidrs, str):
```
HomayoonAlimohammadi, line 216: Shouldn't this be handled on the Juju config level? Like when we define the `type: string` for a config?

HomayoonAlimohammadi (commented):
Did a first pass, great work, thanks a lot @mateoflorido! Some minor comments and questions.

charms/worker/k8s/src/certificates.py
mateoflorido: `config.get()` returns `(bool, int, float, str)`, so this is just a runtime check to ensure that this returns a `str` and to satisfy the static checker.

charms/worker/k8s/charmcraft.yaml
```diff
@@ -215,6 +215,24 @@ config:
...
+    external-certs-domain-name:
+      type: string
+      default: ""
+      description: ""
+    external-certs-apiserver-common-name-format:
+      type: string
+      default: "kube-apiserver"
+      description: |
+        Formatter string for the API server's TLS certificate common name (CN).
+        Supports placeholders: $node_name, $cluster_name, and $domain_name.
```
HomayoonAlimohammadi, line 227: Maybe we can have some examples here as well?

HomayoonAlimohammadi (commented):
Would love someone else to also review this. Maybe @addyess?
Overall, looks good to me. Thanks a lot @mateoflorido!

charms/worker/k8s/src/certificates.py
```diff
@@ -154,50 +217,36 @@ def _certificates_request_mapping(
...
         }
 
-        if self._charm.lead_control_plane:
-            return csr_attributes
-        if self._charm.is_control_plane:
-            return {k: v for k, v in csr_attributes.items() if k in CONTROL_PLANE_CERTIFICATES}
-        return {k: v for k, v in csr_attributes.items() if k in WORKER_CERTIFICATES}
+        try:
+            template = Template(formatter)
+            formatted_name = template.safe_substitute(tmp_context)
```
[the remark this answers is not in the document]
mateoflorido: Also, format can be used to execute arbitrary lookups, which is a security risk 😉

eaudetcobello (approved):
LGTM

charms/worker/k8s/src/certificates.py
```diff
@@ -209,8 +258,13 @@ def _get_service_ips(self) -> Set[str]:
...
         service_ips = set()
-        service_cidrs = self._charm.config.get("bootstrap-service-cidr", "").split(",")
-        for cidr in service_cidrs:
+        service_cidrs = self._charm.config.get("bootstrap-service-cidr", "")
+        if not isinstance(service_cidrs, str):
+            log.warning("Service CIDRs is not str, instead %s", type(service_cidrs))
+            return set()
+        cidrs = service_cidrs.split(",")
+
+        for cidr in cidrs:
```
addyess, line 267: if it's really just a runtime check for the type tester -- can we just use `cast` instead?
```suggestion
        service_cidrs = self._charm.config.get("bootstrap-service-cidr", "")
        cidrs = cast(str, service_cidrs).split(",")

        for cidr in cidrs:
```

charms/worker/k8s/src/utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+import re
+
+# Constants for time calculations
+YEAR_SECONDS = 365 * 24 * 60 * 60  # seconds in a year
+MONTH_SECONDS = 30 * 24 * 60 * 60  # seconds in a month
+DAY_SECONDS = 24 * 60 * 60  # seconds in a day
+
+
+def ttl_to_seconds(ttl: str):
```
addyess, line 14: Does this match some kind of well known format?  Something in my head is going off because this module doesn't include `datetime` or `timedelta` at all.  Help me down from the cliff here.  

I found a few packages that do this kind of datetime math:

https://arrow.readthedocs.io/en/latest/index.html#
https://bear.im/code/parsedatetime/docs/index.html

addyess (approved):
Honestly, this gets a LGTM.  I still have questions/comments but they aren't mandatory other than addressing the missing description on `external-certs-domain-name`

charms/worker/k8s/src/utils.py
mateoflorido: Yes, it's the agreed format in the snap (and pretty much in other parts of Kubernetes upstream). For instance: https://github.com/canonical/k8s-snap/blob/6ee37dbc0124c5099338daae8ced9af24689db7d/src/k8s/pkg/utils/time.go#L43

charms/worker/charmcraft.yaml
```diff
@@ -156,6 +156,14 @@ parts:
...
   pre-upgrade-check:
     description: Run necessary pre-upgrade checks before executing a charm upgrade.
+  refresh-certs:
+    description:  Refresh the certificates of the Kubernetes node
+    params:
+      expires-in:
+        description: Time until the certificates expire, e.g., 1h, 2d, 4mo, 5y
+        type: string
+    required:
+    - expires-in
```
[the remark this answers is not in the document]
mateoflorido: Maybe we can improve the UX. What if we hook it to the update-status event, I recently added an endpoint in the snap to fetch the certificate's validity. I can create a follow up card to surface the certificates status. Wdyt @addyess ?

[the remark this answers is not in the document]
addyess: Sure, a future improvement is fine

PR #397, opened 2025-03-13
test: Add az node labels e2e tests

HomayoonAlimohammadi:
### Overview
This PR adds AZ node labels check to the e2e tests.


HomayoonAlimohammadi:
These tests are not passing since we don't have AZs in the PS6 GH runners. However we have AZs on PS5, at least the `stg-k8s-openstack` env does.
I'm discussing this with IS on MM

tests/integration/test_k8s.py
```diff
@@ -89,15 +89,20 @@ async def test_nodes_labelled(request, kubernetes_cluster: juju.model.Model):
...
+    az_nodes = [n for n in nodes if az_label in n["metadata"]["labels"]]
     assert len(k8s.units + worker.units) == len(labelled), (
         "Not all nodes labelled with custom-label"
     )
     assert len(k8s.units + worker.units) == len(juju_nodes), (
         "Not all nodes labelled as juju-charms"
     )
+    assert len(k8s.units + worker.units) == len(az_nodes), (
+        f"Not all nodes labelled with {az_label}"
+    )
```
addyess, line 105: Maybe you should move these to the openstack tests since `test_k8s` isn't guaranteed to run on a cloud that has AZ support like openstack does

HomayoonAlimohammadi: Correct, that makes sense. For now we can close this PR as we don't run the CI on private runners.
I have a card to add these tests (in the correct place) in the future.

HomayoonAlimohammadi:
Closing this as we're not running the CI on a substrate that supports AZs. We have a card to add these tests in the future.

PR #409, opened 2025-03-21
Select appropiate local charm build now that there are multiple bases…

addyess:
Now that `ops.build_charm(...)` can produce multiple charm files, it's necessary to narrow down which one to use in the integration tests.  

this doesn't affect CI as much as it does local users.


eaudetcobello (approved):
LGTM

PR #417, opened 2025-03-28
fix: use UV in TICS jobs

mateoflorido:
## Overview
Fix the TICS script job


addyess (approved):
Nice, thanks @mateoflorido

PR #425, opened 2025-04-04
feat: add option to accept clouds for integration tests

rapour:
### Overview

This PR suggests adding a custom `clouds` option which enables the user to provide a list of cloud providers for which the respective marked tests will be run. For example, if user provides the `openstack` parameter to the `clouds` option, the tests marked with `openstack` will be run. A few considerations:

1) Only tests that are marked with clouds will be considered for this filter. All other tests will be run irrespective of the cloud option
2) Currently there is no way to exclude all cloud marked tests from the test suite. 

The implication of (1) and (2) is that we if we only have the `openstack` mark for some tests and all the other tests are unmarked for cloud, we cannot drop the `openstack` tests and proceed with others.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -55,6 +55,11 @@ def pytest_addoption(parser: pytest.Parser):
         Apply proxy to model-config.
     --timeout
         Set timeout for tests
+    --clouds
+        Comma separated list of cloud providers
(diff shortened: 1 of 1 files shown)


rapour:
We already have the necessary capability to isolate the tests based on different environments. As a result, after discussions with the team we have decided to close this PR.

PR #441, opened 2025-04-22
fix: change local-storage description

mateoflorido:
### Overview

https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/howto/storage/storage/ says:

> Local storage will not survive node removal, so it may not be suitable for certain setups, such as production multi-node clusters.

whereas https://charmhub.io/k8s/configurations#local-storage-enabled  says (for local-storage-enabled):

> This is useful for development and testing purposes. It is not recommended for production use.

The latter is inconsistent and confusing for users.

### DIFF
--- a/charms/worker/k8s/charmcraft.yaml
+++ b/charms/worker/k8s/charmcraft.yaml
@@ -375,9 +375,9 @@ config:
       default: true
       description: |
         Enable local storage provisioning. This will create a storage class
-        named "local-storage" that uses the hostPath provisioner. This is
-        useful for development and testing purposes. It is not recommended for
-        production use.
+        named "local-storage" that uses the hostPath provisioner. Local storage
+        will not survive node removal, so it may not be suitable for certain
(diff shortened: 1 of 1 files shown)


nhennigan (approved):

charms/worker/k8s/charmcraft.yaml
```diff
@@ -375,9 +375,9 @@ config:
       default: true
       description: |
         Enable local storage provisioning. This will create a storage class
-        named "local-storage" that uses the hostPath provisioner. This is
-        useful for development and testing purposes. It is not recommended for
-        production use.
+        named "local-storage" that uses the hostPath provisioner. Local storage
```
addyess, line 378: it's not really a hostPath provisioner is it?
```suggestion
        named "local-storage" that uses the rawfile-localpv provisioner. Local storage
```

addyess (commented):
dumb nit

PR #446, opened 2025-04-23
The promote-charm action needs the checked-out code for access to metadata

addyess:
Without access to the `charmcraft.yaml`, the promote-charm action cannot perform metadata inspection from the charm file. 

So, we must checkout


mateoflorido (approved):

HomayoonAlimohammadi (approved):
Hopefully we can finally get this job to work. Thanks @addyess! LGTM

PR #463, opened 2025-05-05
fix: remove snaps during removal related events

mateoflorido:
### Overview

The `k8s` and `k8s-worker` charms now remove installed snaps during charm removal.

### Rationale

Previously, installed snaps were not removed when tearing down the unit. This pull request adds support for the `management` method to ensure snaps are removed only during removal related events in Juju.

### Testing
Manual testing:
```
❯ juju remove-application k8s
WARNING This command will perform the following actions:
will remove application k8s
- will remove unit k8s/0

Continue [y/N]? y

Every 2.0s: snap list                                                                juju-1c33ed-1: Mon May  5 20:33:31 2025

Name    Version        Rev    Tracking       Publisher    Notes
core20  20230308       1852   latest/stable  canonical**  base
k8s     v1.33.0        3416   latest/edge    canonical**  classic,held
lxd     5.0.2-838e1b2  24322  5.0/stable/…   canonical**  -
snapd   2.59.1         18933  latest/stable  canonical**  snapd

Every 2.0s: snap list                                                                juju-1c33ed-1: Mon May  5 20:34:54 2025

Name    Version        Rev    Tracking       Publisher    Notes
core20  20230308       1852   latest/stable  canonical**  base
lxd     5.0.2-838e1b2  24322  5.0/stable/…   canonical**  -
snapd   2.59.1         18933  latest/stable  canonical**  snapd
```

### DIFF
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -1048,6 +1048,7 @@ class K8sCharm(ops.CharmBase):
             self._revoke_cluster_tokens(event)
         self.update_status.run()
         self._last_gasp()
+        snap_management(self, remove=True)
 
         relation = self.model.get_relation(CLUSTER_RELATION)
         local_cluster = self.get_cluster_name()
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -1047,8 +1047,8 @@ class K8sCharm(ops.CharmBase):
         if self.lead_control_plane:
             self._revoke_cluster_tokens(event)
         self.update_status.run()
-        self._last_gasp()
-        snap_management(self, remove=True)
+        if self._last_gasp():
+            snap_management(self, remove=True)
 
         relation = self.model.get_relation(CLUSTER_RELATION)
         local_cluster = self.get_cluster_name()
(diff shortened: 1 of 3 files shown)


addyess (approved):
100%

charms/worker/k8s/src/charm.py
```diff
@@ -1048,6 +1048,7 @@ def _death_handler(self, event: ops.EventBase):
             self._revoke_cluster_tokens(event)
         self.update_status.run()
         self._last_gasp()
+        snap_management(self, remove=True)
```
addyess, line 1051: it's possible after last_gasp, the node is still clustered.  

maybe `_last_gasp(...)` should return a value if it succeeded, and only purge when the `_last_gasp(..)` succeeds

mateoflorido: I've updated the code. Wdyt?

charms/worker/k8s/src/charm.py
```diff
@@ -1047,7 +1047,8 @@ def _death_handler(self, event: ops.EventBase):
         if self.lead_control_plane:
             self._revoke_cluster_tokens(event)
         self.update_status.run()
-        self._last_gasp()
+        if self._last_gasp():
+            snap_management(self, remove=True)
```
HomayoonAlimohammadi, line 1051: QQ: Why don't we remove the snap if the unclustering times out?

mateoflorido: We have all the removal events to try this unclustering process (e.g., `stop`, `*-relation-broken` and `remove`). Adam suggested we should only remove them if the unclustering was successful: https://github.com/canonical/k8s-operator/pull/463#discussion_r2080305508

PR #467, opened 2025-05-09
fix: install project deps for TiCS

mateoflorido:
### Overview

Install project dependencies required by the TiCS agent.


addyess (approved):
I can't see anything wrong with this... i don't understand the failures either

PR #485, opened 2025-05-22
Apply charm extra-args at runtime

addyess:
### Overview
* Support run-time adjustment of the extra-args on the kubernetes services
* Adds `node-ip` as a run-time setting based on the ip addresses bound to the `cluster` endpoint binding

---

**Note** 
this PR targets the release branch 1.32.  Accepting it means we'll need to port into 1.33 and main branches


charms/worker/k8s/src/config/arg_files.py
```diff
@@ -0,0 +1,124 @@
...
+            ("kube-scheduler", KUBE_SCHEDULER_ARGS, self.extra_node_kube_scheduler_args),
+            ("kube-proxy", KUBE_PROXY_ARGS, self.extra_node_kube_proxy_args),
+            ("kubelet", KUBELET_ARGS, self.extra_node_kubelet_args),
+        ]:
+            adjusted_services = []
+            if file_path.exists():
+                updated_args = {**self._service_args[service], **extra_args}
+                content, hash_val = self._file_content(updated_args)
+                if hash_val != self._hash[service]:
+                    log.info("Restarting '%s' to adjust %s", service, file_path)
```
mateoflorido, line 121:
```suggestion
                    log.info("Detected changes in args for '%s'. Updating %s and scheduling service restart.")
```

mateoflorido (approved):
Great work, Adam. Just a couple of small suggestions/nits.

PR #500, opened 2025-06-06
fix: Use TIOBE TiCS action

mateoflorido:
Applicable spec: <link>

### Overview

Use the upstream TIOBE action.


addyess (approved):
Thank you thank you @mateoflorido

PR #545, opened 2025-06-27
Move away from nested tox.ini

addyess:
### Overview
* Removes the nested `tox.ini` to speed up testing and formatting

### Details
* deletes the nested tox.ini file
* moves all the logic OUT to the main tox.ini
* maintains as much of the project settings into the nested pyproject.toml


mateoflorido (commented):
Looks great! Quick question, can we remove the `test_requirements.txt` file and just build a dep group with those deps?

Issue #567, opened 2025-07-09
Failing to run the pre-upgrade check needs a better UX


addyess:
### Bug Description

Failing to run the `pre-upgrade-check`, yet running the charm upgrade anyway plays bad UX tricks on the User. 

What's really needed is a way for a newly upgrade charm to detect it's been upgraded without the pre-check being run.  This would block the charm and force the user to downgrade first before upgrading. 

### To Reproduce

```
juju deploy k8s -n 3 --channel=1.32/beta
... wait for stable
juju refresh k8s --channel=latest/edge
```

^^ All units should block with that status:
`blocked: Unit was upgraded without a pre-upgrade-check`

Should be able to downgrade like this
```
juju run k8s/leader pre-upgrade-check
juju refresh k8s --channel=1.32/beta
... wait for stable
```

Then properly upgrade like this
```
juju run k8s/leader pre-upgrade-check
juju refresh k8s --channel=latest/edge
```

### Environment

Any Cloud

### Relevant log output

```shell
n/a
```

addyess:
https://github.com/canonical/mysql-router-operator/  provides an example charm using this library

PR #568, opened 2025-07-09
Improves pre-upgrade-check safely runs before upgrade

addyess:
Addresses #567 

### Overview
* Blocks all the applications if the `pre-upgrade-check` action isn't run before upgrading to this code. 

### Details
* Running the `pre-upgrade-check` action is required to put the charmed application (k8s or k8s-worker) in a state ready to upgrade
* Failing to run this check will upgrade the charm, but not the underlying snap
* These changes block the reconciler from continuing until two things happen:
   1) the `pre-upgrade-check` action is RUN to prepare for a `revert`
   2) the `juju refresh ...` is RUN to revert the a previous version

Once reverted, the admin can properly execute the `pre-upgrade-check` and `juju refresh` back to a version that includes this code.

This is accomplished in the `K8sUpgrade.handler(...)` method:
1) Allowing `UpgradeGranted` and `UpgradeFinished` custom-events from the upgrade library to continue through the reconciler.  
2) Allowing an uninitialized `upgrade.state` (`None`) to pass to the reconciler
3) Retrying a failed unit if it's in the `upgrade.state` -- but prohibiting the reconciler if it fails
4) Prohibited the reconciler if the `upgrade.state` is not idle
5) watching for an `upgrade-charm` event and reacting like this:
    1) if it's a charm resource attachment -- pass to the reconciler
    2) if its a single unit application and the upgrade was already completed in the `upgrade-charm` handler in the library then pass to the reconciler
    3) If the upgrade-stack is populated, we know the pre-upgrade check WAS run -- so continue to the reconciler
    4) Otherwise, the user didn't run the `pre-upgrade-check`: block the charm, fail the upgrade


charms/worker/k8s/src/charm.py
```diff
@@ -998,6 +999,32 @@ def _death_handler(self, event: ops.EventBase):
...
+            log.info("Upgrade finished, proceeding with reconciliation")
+            return
+        if isinstance(event, ops.UpgradeCharmEvent) and not self.upgrade.upgrade_stack:
+            message = "Unit was upgraded without a pre-upgrade-check"
+            log.warning(message)
+            self.upgrade.set_unit_failed(message)
+            if self.unit.is_leader():
+                self.upgrade.upgrade_stack = self.upgrade.build_upgrade_stack()
+            status.add(self.unit.status)
+            raise ReconcilerError(message)
```
addyess, line 1021: oh no @mateoflorido ... does attaching a resource triggers an `update-charm` event!!

charms/worker/k8s/src/upgrade.py
```diff
@@ -188,41 +204,34 @@ def _on_upgrade_granted(self, event: UpgradeGrantedEvent) -> None:
...
-            services = list(K8S_CONTROL_PLANE_SERVICES)
-            bootstrap_datastore = BOOTSTRAP_DATASTORE.get(self.charm)
-            if bootstrap_datastore != DATASTORE_TYPE_K8S_DQLITE:
-                services.remove(K8S_DQLITE_SERVICE)
-            if bootstrap_datastore != DATASTORE_TYPE_ETCD:
-                services.remove(MANAGED_ETCD_SERVICE)
-        else:
-            services = list(K8S_WORKER_SERVICES)
+        snap_services = list_services(SNAP_NAME)
+        services = sorted(name for name, service in snap_services.items() if service["enabled"])
```
addyess, line 231: This is now quite an important change. 

rather than use the charm config to determine which services should be stopped/restarted after the upgrade -- we need to use the current set of snap services on the machine that are running

One reason for this is if we're upgrading a cluster bootstrapped with `k8s-dqlite`  -- on upgrade it could be configured for `managed-etcd`.  We'd fail to stop the dqlite service on upgrade here b/c we're looking at charm config rather than actual running services.   

Let's instead restart the `enabled` services of the k8s snap.

charms/worker/k8s/src/upgrade.py
```diff
@@ -232,24 +241,22 @@ def _upgrade(self, event: Union[ops.EventBase, ops.HookEvent]) -> None:
...
-        if not relation:
-            return [int(self.charm.unit.name.split("/")[-1])]
-
-        return [
-            int(unit.name.split("/")[-1]) for unit in ({self.charm.unit} | set(relation.units))
-        ]
+        units = {self.charm.unit}
+        if relation := self.charm.model.get_relation(UPGRADE_RELATION):
+            units |= set(relation.units)
+        return [int(unit.name.split("/")[-1]) for unit in units]
```
addyess, line 259: golf

charms/worker/k8s/src/upgrade.py
```diff
@@ -68,25 +63,52 @@ def __init__(self, charm: K8sCharmProtocol, cluster_inspector: ClusterInspector,
...
+            elif super().upgrade_stack:
+                log.debug("Upgrade stack exists, proceed with reconciliation.")
+            else:
+                reason = "Unit was upgraded without a pre-upgrade-check"
+                log.warning(reason)
+                self.set_unit_failed(reason)
+                if self.charm.unit.is_leader():
+                    self.upgrade_stack = self.build_upgrade_stack()
+                status.add(self.charm.unit.status)
+                raise status.ReconcilerError(reason)
```
addyess, line 111: The bulk of the logic really went into this one function and is documented in the PR description

mateoflorido (approved):
Great work, Adam! Thanks for improving the upgrade handler, this looks much cleaner.

mateoflorido:
@addyess Do we need to backport these changes to the release branches?

addyess:
> @addyess Do we need to backport these changes to the release branches?

We could but I'm not sure it's necessary

PR #586, opened 2025-07-22
Catch a broader httpx exception to signal a failure cluster inspection

addyess:
### Overview
* `httpx.ConnectError` is too narrow of an issue to catch for cluster inspection.  It should more use the more broad exception `httpx.HTTPError`.  if anything goes wrong checking the api server, that should be noted as a `ClusterInspectorError`. 

### Details
* seen in an integration tests run where the client through another error, and the charm went into ERROR state.


mateoflorido (approved):
LGTM. For anyone curious (or future reference), here's the hierarchy of HTTPX exceptions: https://www.python-httpx.org/exceptions/

PR #590, opened 2025-07-23
fix: await public ip address to confirm its not the LB ip address

addyess:
### Overview
* need to await a juju method to get the pubilc ip address of the openstack instance to compare against the load-balancer address

### Details
* this previously produced a warning but not a test failure.

````
tests/integration/test_openstack.py::test_external_load_balancer
  /home/ubuntu/actions-runner/_work/sqa-cloud-deployment-pipeline/sqa-cloud-deployment-pipeline/canonicalk8s_test_openstack/tests/integration/test_openstack.py:86: RuntimeWarning: coroutine 'Unit.get_***_address' was never awaited
    assert unit.get_***_address() != server_endpoint, "External lb not configured"
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.

During handling of the above exception, another exception occurred:
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

### DIFF
--- a/tests/integration/terraform/default-manifest.yaml
+++ b/tests/integration/terraform/default-manifest.yaml
@@ -2,12 +2,12 @@
 # See LICENSE file for licensing details.
 
 k8s:
-    units: 3
-    base: ubuntu@24.04
-    constraints: arch=amd64 cores=2 mem=4096M root-disk=16384M
-    channel: latest/edge
-k8s_worker:
-    units: 2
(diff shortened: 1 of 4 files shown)


mateoflorido (approved):
Amazing, I see it now 😄. LGTM

PR #594, opened 2025-07-25
fix: overshare datastore-extra-args since they're scoped in the k8s-snap

addyess:
Applicable spec: KU182

### Overview

Rather than depending on a `bootstrap-datastore` config the gate whether to share to config to the snap, the snap will use the extra args as its sees fit.  

### Rationale

This drops a dependency in the code where we check `bootstrap-datastore`.  It allows the bootstrap preventer logic to be a "read-only" status issue and not a reconciler blocker.

### Juju Events Changes

None

### Module Changes

* Happens to also extend the `FileConfig` feature to allow changing control-plane arguments in real-time with charm config.

### Library Changes

None

### Checklist


charms/worker/k8s/src/config/extra_args.py
```diff
@@ -18,13 +19,18 @@
 )
 
 
-def _parse(config_data) -> Dict[str, str]:
+def _parse(option: StrOption, charm: ops.CharmBase) -> Dict[str, str]:
```
mateoflorido, line 22: Can we target the super class `CharmOption`?

mateoflorido (approved):
LGTM. Just a nit below

charms/worker/k8s/src/config/extra_args.py
addyess: um, if we do, the type of the value in `option.get(charm)` wouldn't be `str` -- it'd be an `Any`.

charms/worker/k8s/src/certificates.py
```diff
@@ -335,26 +334,79 @@ def _populate_bootstrap_certificates(self, bootstrap_config: BootstrapConfig) ->
...
-
-        # NOTE: This operation is safe because we're validating the provider during the
-        # certificate configuration in the `configure_certificates` method.
         relation = self.model.get_relation(CLUSTER_RELATION)
-        if not relation:
-            return None
+        provider = relation and relation.data[relation.app].get(CLUSTER_CERTIFICATES_KEY)
+        if not provider and self._charm.is_control_plane:
+            provider = BOOTSTRAP_CERTIFICATES.get(self._charm)
+            self._validate_provider(provider)
```
addyess, line 347: until it's persisted, keep using the charm config

deuberger:
@addyess how does this related to KU-3838 (https://github.com/canonical/k8s-operator/issues/567) or was the branch name a mistake?

PR #598, opened 2025-07-29
Ku 3939/overshare datastore args/cp bootstrap or join

addyess:
Applicable spec: KU182
Replaces #594 with adjusted branch name

### Overview

* Protect `bootstrap-datastore` from use except during bootstrap
* Protect `bootstrap-certificates` from use except during bootstrap

### Rationale

This isolates reading charm config of  `bootstrap-datastore` and `bootstrap-certificates` to only during bootstrap.
It allows the bootstrap preventer logic to be a "read-only" status issue and not a reconcile blocker.

### Juju Events Changes

None

### Module Changes

* `charm`
    * Never use the charm configured `bootstrap-datastore` except when assembling the bootstrap config.
    * Never use the charm configured `bootstrap-certificates` except when assembling the bootstrap config.
    * Once bootstrapped, ONLY use the current datastore type from ClusterConfig
    * Once bootstrapped, ONLY use the cluster relation app databag key for the provider.
* `config.arg_files`
    * extends the `FileConfig` feature to allow changing control-plane arguments in real-time with charm config.
* `certificates`
    * isolate the `bootstrap-certificates` to only be referenced during bootstrap

### Library Changes

None

### Checklist


charms/worker/k8s/charmcraft.yaml
```diff
@@ -93,7 +93,7 @@ config:
       description: |
         Space-separated list of taints to apply to this node at registration time.
 
-        This config is only used at bootstrap time when Kubelet first registers the
+        This config is only used at node join time when Kubelet first registers the
```
deuberger, line 96: This could be used at bootstrap for the initial k8s charm too, right? Maybe "This config is only used when Kubelet first registers the node with Kubernetes (bootstrap or node join)," would more clear?

charms/worker/k8s/src/certificates.py
```diff
@@ -335,26 +337,90 @@ def _populate_bootstrap_certificates(self, bootstrap_config: BootstrapConfig) ->
...
-            log.info("Waiting for certificates provider")
-            return None
+        if not provider:
+            # Note(AKD): This could be an upgrade scenario where the provider is unset.
+            # if this node is online, we know that a certificates provider is already set.
+            try:
+                self._charm.api_manager.get_node_status().metadata
+                provider = "self-signed" if pki.check_ca_key() else "external"
+            except (K8sdConnectionError, InvalidResponseError) as e:
+                log.error("Failed to get node status: %s", e)
```
mateoflorido, line 357: Should we raise here and prevent the charm to keep going with an `None` value? I see the logic blocks while validating the provider, but I'd prefer a different approach, as the status `Invalid certificate issuer: ''` could be misleading.

mateoflorido (commented):
LGTM, just a few minor comments below:

charms/worker/k8s/src/certificates.py
```diff
@@ -335,26 +337,90 @@ def _populate_bootstrap_certificates(self, bootstrap_config: BootstrapConfig) ->
...
 
-        if not (provider := relation.data[relation.app].get(CLUSTER_CERTIFICATES_KEY)):
-            log.info("Waiting for certificates provider")
-            return None
+        if not provider:
+            # Note(AKD): This could be an upgrade scenario where the provider is unset.
+            # if this node is online, we know that a certificates provider is already set.
+            try:
+                self._charm.api_manager.get_node_status().metadata
+                provider = "self-signed" if pki.check_ca_key() else "external"
```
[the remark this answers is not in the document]
mateoflorido: New API: https://documentation.ubuntu.com/canonical-kubernetes/latest/snap/reference/commands/#k8s-certs-status

charms/worker/k8s/src/certificates.py
addyess: I think i went down that chain at first, but there's this chicken/egg issue with the `K8sCertificates` class is used outside the reconciler, so raising here wouldn't be good..

```
K8sCharm.__init__(...)
  K8sCertificates.__init__(...)
    K8sCertificates._init_certificates_provider(...)
       K8sCertificates.get_provider_name(...)
... HERE ...
```

addyess: also, this error you mentioned wouldn't show up unless the config option itself is set to `""`

```python
        if not provider and self._charm.is_control_plane:
            provider = BOOTSTRAP_CERTIFICATES.get(self._charm)
            self._validate_provider(provider)
```

I should deploy the charm with these changes with `juju deploy k8s --config bootstrap-certificates=""` and see how this is handled.  We'll likely end up in an error state.

charms/worker/k8s/src/certificates.py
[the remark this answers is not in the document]
addyess: There is no guarantee that the snap is even installed when this is first called.  This could be issued pre-bootstrap -- hence the call first to `get_node_status` 

but, supposing the node is bootstrapped, it's still possible its running the k8s 1.32 snap (an upgrade scenario).  So -- there's no guarantee that API exists yet.

deuberger:
@addyess I think the description for this PR is missing: "Updates the bootstrap-node-taints description"

PR #607, opened 2025-08-01
fix: Regenerate tokens when node-join fails

addyess:
### Overview

* joining nodes need to feedback a join failure to the lead k8s unit when its allocated token fails to join the cluster
* lead k8s unit will regenerate a new secret and watch for failures of the new token when a joiner reports a failure

### Details
* create 2 new pydantic models ( `TokenFailure` and `TokenContent`) for json serialization/deserialization
* `TokenFailure` represents the error encountered when using a revision of a  join-token and a failure occurs
* `TokenContent` represents a join-token and its atomically increasing revision number. This content is pushed into a juju secret
* replaces a few "magic-strings" with string literals
* catches any error trying to use a join-token, and sharing it over the `cluster` relation in the `token-failure` field
* clears any `token-failure` field when the unit confirms its in the cluster
* lead-control-plane units watch for a `token-failure` with a revision matching the current token-content it's sharing with the joiner
    * if the revision's match, the joiner is declaring it's failed to join with the provided token
    * a new token is generated (with a new revision number)
    * this triggers the joiner to try again with a new join-token


charms/worker/k8s/src/token_distributor.py
```diff
@@ -24,10 +31,92 @@
...
+
+    @field_serializer("revision", when_used="always")
+    def _revision_as_str(self, v: int) -> str:
+        """Convert revision to str for serialization."""
+        return str(v)
+
+    @field_serializer("token", when_used="always")
+    def _token_as_str(self, v: SecretStr) -> str:
+        """Expose token to str for serialization."""
+        return v.get_secret_value()
```
addyess, line 81: this is the pile of serializer/deserializers to make sure that dict or json model dumps always create a `dict[str,str]` for juju secrets to use

addyess (commented):
Document the PR

addyess:
@mateoflorido 

> OHHH, I see it now. We need to make sure everything is a dict[str, str] because of the Juju secrets API, right?


right juju secrets must be `dict[str,str]`, but I need an `int` in order to create a monotonically increasing revision and to compare with the `token-failure`.  So, i'm using pydantic here to create a TokenContent object that always serializes to `dict[str, str]` but deserializes the revision into an int

PR #616, opened 2025-08-11
fix: Regenerate tokens when node-join fails

addyess:
Backports: #607 

### Overview
* Regenerate tokens when node-join fails

### Details
* Ports changes from #607 to match pydantic 1 style invocation


bschimke95 (approved):
Great! Fingers crossed that this unblocks the release.

PR #618, opened 2025-08-12
feat: add `requires` etcd terraform output 

mateoflorido:
### Overview

Add the requires endpoints for the Charmed Etcd integration in Terraform


addyess (approved):
yep!  thanks @mateoflorido

PR #623, opened 2025-08-14
Do not manage service restarts when performing a snap refresh

addyess:
Applicable spec: KU-182

### Overview
The snap should now be able to handle restarting its own services, this shouldn't be in the context of the charm.  

### Rationale

This was only ever here in order to workaround some issues with k8s 1.31, and it's masking some bad decisions.  A snap refresh should complete on its own without having to stop/start the k8s services.  It can badly impact quorum on applications which perform raft -- so we just need to not attempt restarts in the charm. 

- [ ] Backported to 1.32

### DIFF
--- a/charms/worker/k8s/src/snap.py
+++ b/charms/worker/k8s/src/snap.py
@@ -13,7 +13,7 @@ import shutil
 import subprocess
 import tarfile
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


mateoflorido (approved):
Great work! LGTM 🌟

PR #630, opened 2025-08-19
fix: address vale errors

addyess:
Address newly discovered vale errors. 

Thanks https://github.com/canonical/operator-workflows/pull/740


nhennigan (approved):
Thanks for resolving these issues!

PR #633, opened 2025-08-21
feat: add a k8s-inspect juju crashdump addon

addyess:
Adds a k8s-inspect addon for dumping the k8s inspect report into the juju crashdump


mateoflorido (approved):
LGTM

PR #647, opened 2025-08-25
[Release 1.33] Update K8s revisions ["amd64-4176","arm64-4180"]

addyess:
Updates K8s revisions for 1.33
* amd64-4176\n *arm64-4180


addyess:
🚧 Don't merge.  Instead use #645 🚧

Issue #648, opened 2025-08-27
cos-tokens / cos-worker-tokens relation breaks if k8s leader unit is deleted


drencrom:
### Bug Description

At first deploy time the k8s leader unit creates the auth tokens for the grafana-agents and shares them over its UNIT databag in the `cos-worker-tokens` and `cos-tokens` relations. When the workers (and other control-planes) look to find their tokens, they walk the unit-data bag, to find the secret-ids with their unit-names, and pull the secret-id.  Using the secret-id, each unit can fetch it's token. They then respond on the relation with a `"joined: <my node-name>"` in their own unit-data, and the k8s leader, seeing that field, marks that the k8s-worker (and k8s control-plane) unit has `joined` (no longer pending)

If at a later time someone removes the k8s leader unit, the leadership changes to a new node, but the previous k8s leader unit data in the `cos-worker-tokens` and `cos-tokens` relation is lost. When this happens the k8s-worker and k8s units still think they are "joined" since both the app data in the relation and their own unit data in the relation confirms it. The problem is that the secret-ids are now gone.  When this happens grafana-agent is updated with erroneous scrape metrics. 

### To Reproduce

1. Deploy a canonical kubernetes cluster 
2. Add relations between k8s nodes (control plane and workers) and grafana-agent units and link them to a COS model via cross model relations
3. Verify that kubernetes metrics are available on COS grafana dashboards
4. Add a 4th k8s control-plane unit
5. Remove the k8s leader unit
6. Verify that kubernetes metrics are not longer seen on grafana dashboards
### Environment

Running canonical k8s 1.32/stable on juju 3.6.
k8s on version 1050 and k8s-worker on version 1052 

### Relevant log output

```shell
Seeing message 'Failed to get COS' on /var/log/juju/unit-k8s* files on the k8s leader unit.
```

addyess:
Closed with release to 1.32/stable

PR #649, opened 2025-08-27
[COS] fix: cos-tokens shared in app databag

addyess:
### Overview
Move get/grant/revoke to the `TokenManager` class to control in which relation buckets secret ids appear.
Move all `secret-ids` to the app databags rather than unit databags to handle losing a k8s leader unit which may be holding the reference id to a juju secret.

### Issues
* #648 Fixes an issue with sharing cos-tokens for grafana-agent metrics scraping

### Backport
* Will need to backport to 1.32 and 1.33 branches


mateoflorido (approved):
Great work, LGTM!

mateoflorido:
fixes #648

PR #658, opened 2025-09-01
feat: Add `managed-etcd` observability

mateoflorido:
### Overview

Introduce support for configuring COS observability for the `managed-etcd` service in `k8s-operator` charms.

### Rationale

The changes enable observability and monitoring of the managed `etcd` service. The main changes include updating argument handling for datastores and the addition of metrics endpoint configuration for managed etcd. Aditionally, some type annotations have been updated from the invariant `Dict` to `Mapping`, allowing for covariant types in certain cases.

### DIFF
--- a/charms/worker/k8s/src/charm.py
+++ b/charms/worker/k8s/src/charm.py
@@ -480,7 +480,8 @@ class K8sCharm(ops.CharmBase):
         bootstrap_config.extra_sans = self._get_extra_sans()
         cluster_name = self.get_cluster_name()
         node_ips = self._get_node_ips()
-        config.extra_args.craft(self.config, bootstrap_config, cluster_name, node_ips)
+        datastore = self.bootstrap.config.datastore
+        config.extra_args.craft(self.config, bootstrap_config, cluster_name, node_ips, datastore)
         return bootstrap_config
 
     def _configure_external_load_balancer(self) -> None:
@@ -771,7 +772,10 @@ class K8sCharm(ops.CharmBase):
         try:
             with self.collector.recover_token(relation) as token:
                 return self.cos.get_metrics_endpoints(
-                    self.get_node_name(), token, self.is_control_plane
+                    self.get_node_name(),
+                    token,
+                    self.is_control_plane,
+                    self.bootstrap.config.datastore,
                 )
         except ReconcilerError:
             log.exception("Failed to get COS token.")
@@ -1064,7 +1068,10 @@ class K8sCharm(ops.CharmBase):
             status.add(ops.MaintenanceStatus("Ensuring Kubernetes Extra Args"))
             file_args_config = config.arg_files.FileArgsConfig()
             node_ips = self._get_node_ips()
-            config.extra_args.craft(self.config, file_args_config, cluster_name, node_ips)
+            datastore = self.bootstrap.config.datastore if self.is_control_plane else None
+            config.extra_args.craft(
+                self.config, file_args_config, cluster_name, node_ips, datastore
+            )
             file_args_config.ensure()
 
     @property
--- a/charms/worker/k8s/src/config/extra_args.py
+++ b/charms/worker/k8s/src/config/extra_args.py
@@ -5,7 +5,7 @@
 
 """Parse extra arguments for Kubernetes components."""
 
-from typing import Dict, List, Union
+from typing import Dict, List, Optional, Union
 
 import literals
 import ops
@@ -35,11 +35,23 @@ def _parse(config_data) -> Dict[str, str]:
     return args
 
 
+def _configure_datastore_args(dest, src, datastore: str):
+    """Configure Service Arguments for the specified datastore."""
+    datastore_args = _parse(src["datastore-extra-args"])
+    if datastore == literals.DATASTORE_TYPE_K8S_DQLITE:
+        dest.extra_node_k8s_dqlite_args = datastore_args
+    elif datastore == literals.DATASTORE_TYPE_ETCD:
+        # NOTE: Enable the metrics URL on localhost.
+        datastore_args[literals.LISTEN_METRICS_URLS_ARG] = "http://localhost:2381"
+        dest.extra_node_etcd_args = datastore_args
+
+
 def craft(
     src: ops.ConfigData,
     dest: Union[BootstrapConfig, ControlPlaneNodeJoinConfig, FileArgsConfig, NodeJoinConfig],
     cluster_name: str,
     node_ips: List[str],
+    datastore: Optional[str] = None,
 ):
     """Set extra arguments for Kubernetes components based on the provided configuration.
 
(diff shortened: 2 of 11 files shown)


charms/worker/k8s/src/config/extra_args.py
```diff
@@ -58,37 +70,35 @@ def craft(
...
-    cmd = _parse(src["kubelet-extra-args"])
-    if node_ips:
-        cmd.update(**{"--node-ip": ",".join(node_ips)})
-    else:
-        cmd.pop("--node-ip", None)
-    dest.extra_node_kubelet_args = cmd
+        controller_args = {"--cluster-name": cluster_name} if cluster_name else {}
+        dest.extra_node_kube_controller_manager_args = update_args(
+            "kube-controller-manager-extra-args", controller_args
+        )
```
addyess, line 88: i think there's a regression here.  We don't want the users to set the `--cluster-name` field.  The charm sets it -- or the charm prevents it

charms/worker/k8s/src/charm.py
```diff
@@ -1064,7 +1068,10 @@ def _apply_extra_args(self):
             status.add(ops.MaintenanceStatus("Ensuring Kubernetes Extra Args"))
             file_args_config = config.arg_files.FileArgsConfig()
             node_ips = self._get_node_ips()
-            config.extra_args.craft(self.config, file_args_config, cluster_name, node_ips)
+            datastore = self.bootstrap.config.datastore if self.is_control_plane else None
```
addyess, line 1071: You had something that's overkill.  `self.bootstrap.config.datastore` is already defined as an `Optional[str]` and will be None on the worker units.

```suggestion
            datastore = self.bootstrap.config.datastore
```

charms/worker/k8s/src/charm.py
```diff
@@ -480,7 +480,8 @@ def _assemble_bootstrap_config(self):
         bootstrap_config.extra_sans = self._get_extra_sans()
         cluster_name = self.get_cluster_name()
         node_ips = self._get_node_ips()
-        config.extra_args.craft(self.config, bootstrap_config, cluster_name, node_ips)
+        datastore = self.bootstrap.config.datastore
+        config.extra_args.craft(self.config, bootstrap_config, cluster_name, node_ips, datastore)
```
addyess, line 484: don't we also need this in [`_join_with_token(...)`](https://github.com/canonical/k8s-operator/pull/658/files#diff-5ba1576e3daac57729756610a1f51b9706129e66f250d7a8557ea5666d0c490aR920)

charms/worker/k8s/src/config/extra_args.py
```diff
@@ -35,11 +35,23 @@ def _parse(config_data) -> Dict[str, str]:
...
 
+def _configure_datastore_args(dest, src, datastore: str):
+    """Configure Service Arguments for the specified datastore."""
+    datastore_args = _parse(src["datastore-extra-args"])
+    if datastore == literals.DATASTORE_TYPE_K8S_DQLITE:
+        dest.extra_node_k8s_dqlite_args = datastore_args
+    elif datastore == literals.DATASTORE_TYPE_ETCD:
+        # NOTE: Enable the metrics URL on localhost.
+        datastore_args[literals.LISTEN_METRICS_URLS_ARG] = "http://localhost:2381"
+        dest.extra_node_etcd_args = datastore_args
```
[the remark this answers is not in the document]
mateoflorido: Even though we have mechanisms to prevent writing arg files if they don’t exist, I think using the same arguments for both datastore configurations could make debugging more difficult later. I’d prefer to be more explicit when configuring a specific datastore, wdyt @addyess ?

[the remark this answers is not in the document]
addyess: sure, its fine.  You can leave this as is

charms/worker/k8s/src/config/extra_args.py
```diff
@@ -35,11 +35,23 @@ def _parse(config_data) -> Dict[str, str]:
...
 
 
+def _configure_datastore_args(dest, src, datastore: str):
+    """Configure Service Arguments for the specified datastore."""
+    datastore_args = _parse(src["datastore-extra-args"])
+    if datastore == literals.DATASTORE_TYPE_K8S_DQLITE:
+        dest.extra_node_k8s_dqlite_args = datastore_args
+    elif datastore == literals.DATASTORE_TYPE_ETCD:
+        # NOTE: Enable the metrics URL on localhost.
+        datastore_args[literals.LISTEN_METRICS_URLS_ARG] = "http://localhost:2381"
```
addyess, line 45: does `http://localhost:2381` need to be a literal at all?

charms/worker/k8s/src/config/extra_args.py
```diff
@@ -58,37 +70,34 @@ def craft(
             The configuration object to be updated with extra arguments.
         cluster_name (str): the name of the cluster to override in the extra arguments.
         node_ips (list[str]): the IP address of the node to override in the extra arguments.
+        datastore(Optional[str]): the name of the datastore.
```
addyess, line 73: its not in the snap form, it's in the charm form right?

```suggestion
        datastore(Optional[str]): the name of the datastore in charm notation
```

charms/worker/k8s/src/config/extra_args.py
```diff
@@ -58,37 +70,34 @@ def craft(
...
-        dest.extra_node_kube_controller_manager_args = cmd
+            args.pop("--cluster-name", None)
+        dest.extra_node_kube_controller_manager_args = args
+
+        dest.extra_node_kube_scheduler_args = _parse(src["kube-scheduler-extra-args"])
 
-        cmd = _parse(src["kube-scheduler-extra-args"])
-        dest.extra_node_kube_scheduler_args = cmd
+        if datastore:
+            _configure_datastore_args(dest, src, datastore)
```
addyess, line 88: what if you push the `if datastore:` into `_configure_datastore_args(...)`

it's already checking the value of datastore passed in...  if the value is `None`, i don't think `dest` will be updated (however src["datastore-extra-args"] would be read.

charms/worker/k8s/src/config/extra_args.py
```diff
@@ -58,37 +70,34 @@ def craft(
...
 
-        cmd = _parse(src["datastore-extra-args"])
-        match src["bootstrap-datastore"]:
-            case literals.DATASTORE_TYPE_K8S_DQLITE:
-                dest.extra_node_k8s_dqlite_args = cmd
-            case literals.DATASTORE_TYPE_ETCD:
-                dest.extra_node_etcd_args = cmd
+    if datastore and isinstance(dest, FileArgsConfig):
+        _configure_datastore_args(dest, src, datastore)
```
addyess, line 91: same here:

```suggestion
    if isinstance(dest, FileArgsConfig):
        _configure_datastore_args(dest, src, datastore)
```

addyess (commented):
Looking good... i just have a few nits now

PR #666, opened 2025-09-05
feat(backport 1.32): Add `managed-etcd` observability 

mateoflorido:
## Overview

Backport #658 --> Introduce support for configuring COS observability for the managed-etcd service in k8s-operator charms.


addyess (approved):
Looks, great.  This is going to be an excellent addition

PR #715, opened 2025-10-16
Repair failing test-cos

addyess:
Adjusts the return type from executing a command on lxd to always yield `(int, bytes, bytes)`

### DIFF
--- a/charms/worker/k8s/src/token_distributor.py
+++ b/charms/worker/k8s/src/token_distributor.py
@@ -299,7 +299,7 @@ class ClusterTokenManager(TokenManager):
             K8sdConnectionError: reraises cluster token remove failures
         """
         try:
-            self.api_manager.remove_node(name)
+            self.api_manager.remove_node(name, force=ignore_errors)
         except (K8sdConnectionError, InvalidResponseError) as e:
             if ignore_errors or getattr(e, "code") == ErrorCodes.STATUS_NODE_UNAVAILABLE:
                 # Let's just ignore some of these expected errors:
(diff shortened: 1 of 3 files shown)


charms/worker/k8s/src/token_distributor.py
```diff
@@ -299,7 +299,7 @@ def remove(self, name: str, secret: Optional[ops.Secret], ignore_errors: bool):
             K8sdConnectionError: reraises cluster token remove failures
         """
         try:
-            self.api_manager.remove_node(name)
+            self.api_manager.remove_node(name, force=ignore_errors)
```
addyess: we only need to force if we intentionally are ignoring errors

tests/integration/cos_substrate.py
```diff
@@ -241,24 +242,24 @@ def execute_command(self, container, command: List[str]):
...
+            rc, stdout, stderr = container.execute(command, decode=False)
             if rc != 0:
                 log.error(
                     "Failed to run %s with return code %s. stdout: %s, stderr: %s",
-                    command,
+                    _cmd,
                     rc,
                     stdout,
                     stderr,
                 )
```
mateoflorido: In this case, we have to decode the stdout and stderr, right?

addyess: by using `decode=False` -- it forces the output/errors to bytes.  Without this -- we were floating based on what the library set as the default.  It turns out it changed (and we were depending on `decode=True`) for everything to work

PR #718, opened 2025-10-21
ARM tests on noble containers

addyess:
### Overview

* Address issues where snap installation on jammy contianers running on noble machines don't cooperate
* Address changes to juju where the lxd profile name must first change https://github.com/juju/juju/pull/20357

### Details
* by running ARM tests on a noble gh runner with  noble containers, we expand our noble testing support
* swapping the lxd profile name from `juju-<model>` to `juju-<model>-<model-uuid[:6]>` we still apply the profile for kubernetes to operate
* addition of a new test called `test_smoke` that is a single node k8s (to speed up testing)

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -76,7 +76,6 @@ jobs:
       juju-channel: 3/stable
       load-test-enabled: false
       provider: lxd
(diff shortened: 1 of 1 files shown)


mateoflorido (approved):
LGTM. Thanks for the fix.

PR #727, opened 2025-10-24
fix: COS Kubernetes platform should match the OS of the runner

addyess:
### Overview
Match the test runner OS to the OS release hosting the cos cluster


mateoflorido (approved):
LGTM. Thanks @addyess!

PR #737, opened 2025-10-29
chore: Replace snap and tls_certificates library from pypi

addyess:
### Overview
Rather than updating charmlibs (#691) , use the libraries from pypi


charms/worker/k8s/src/certificates.py
```diff
@@ -263,16 +263,23 @@ def _get_kubernetes_svc_ips(self) -> Set[str]:
...
         Returns:
             Tuple[ProviderCertificate, PrivateKey]: A tuple containing the certificate and key.
 
         Raises:
             ReconcilerError: If the certificate/key pair is missing.
         """
+        if not request:
+            self._refresh_event.emit()
+            raise status.ReconcilerError("Invalid certificate request.")
+
```
addyess: new mypy linting fix

PR #754, opened 2025-11-19
fix: handle node already in cluster error during join operation

mateoflorido:
## Overview
Handle node already in cluster error during join operation

## Rationale

When a node attempts to join a cluster but is already part of it (k8sd returns the error code 521), the charm now handles this case by logging and continuing instead of failing. This makes the join operation idempotent and more resilient to transient states.

The error code 521 (STATUS_NODE_IN_USE) is defined in the k8s snap:
https://github.com/canonical/k8s-snap/blob/6f1adf57f26719573e92cf5074ac7d89d4e97203/src/k8s/pkg/k8sd/api/response.go#L10-L11

Also, after seeing the charm "overgranting" the secret after these type of errors (i.e., charmed-etcd rebase for [amd64](https://github.com/canonical/k8s-operator/actions/runs/19472674167/job/55727304217) and [arm64](https://github.com/canonical/k8s-operator/actions/runs/19472674167/job/55725783353)) . We have decided to remove these grant calls, and just let the charm use the already granted secret.


charms/worker/k8s/src/token_distributor.py
```diff
@@ -612,8 +612,6 @@ def allocate_tokens(  # noqa: C901
                 self.update_node(relation, unit, f"joined-{node}")
                 if tokenizer.revoke_on_join:
                     tokenizer.revoke(relation, self.charm, unit)
-                elif secret:
```
bschimke95: Is this change related to this PR?

addyess (approved):
I'm glad we agreed on this solution together.

Issue #755, opened 2025-11-20
The containerd-custom-registries configuration applies only to the control-plane units


chrisbetze:
### Bug Description

Trying to set the [containerd-custom-registries](https://charmhub.io/k8s/configurations#containerd-custom-registries) config to allow containerd to pull images from registries where auth is required, I noticed that the changes are reflected only in the control-plane units and not in the workers.

I cannot see a similar configuration option on the [k8s-worker](https://charmhub.io/k8s-worker/configurations) charm.

### To Reproduce

Setting the `containerd-custom-registries` config:
```
juju config k8s containerd-custom-registries='[{"url":"https://registry-1.docker.io","host":"docker.io","username":"username","password":"password"}]'
```

The related `/etc/containerd/hosts.d/docker.io/hosts.toml` file will be created only on the control-plane units and not on the worker units.

### Environment

Tested it on `1.32/stable: 987`.

### Relevant log output

```shell
N/A
```

PR #758, opened 2025-11-21
docs: link to upstream documentation page

addyess:
## Description

Improve the documentation around the containerd relation according to 
https://github.com/canonical/k8s-operator/issues/755

## Solution

Link to upstream docs of this feature so its more visible when looking at charmhub

## Issue

https://github.com/canonical/k8s-operator/issues/755

### DIFF
--- a/charms/worker/k8s/charmcraft.yaml
+++ b/charms/worker/k8s/charmcraft.yaml
@@ -486,8 +486,3 @@ requires:
     interface: gcp-integration
   external-load-balancer:
     interface: loadbalancer
-  etcd-certificates:
-    interface: tls-certificates
-    limit: 1
-  etcd-client:
-    interface: etcd_client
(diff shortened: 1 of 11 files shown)


charms/worker/k8s/charmcraft.yaml
```diff
@@ -125,6 +125,8 @@ config:
         Registry endpoints and credentials. Setting this config allows containerd
         to pull images from registries where auth is required.
 
+        https://documentation.ubuntu.com/canonical-kubernetes/latest/charm/howto/custom-registry/
```
nhennigan: Are we happy to link to latest always? Should this be versioned?

nhennigan (approved):

charms/worker/k8s/charmcraft.yaml
nhennigan: Fair enough. :)

Issue #762, opened 2025-11-25
Documentation bug: Typo in datastore-extra-args config description


vishvikkrishnan:
The docstring has a typo and is referring to kube-controller-manager instead of the internal datastore (managed-etcd/dqlite).

The [typo is seen](https://charmhub.io/k8s/configurations#datastore-extra-args:~:text=2380%20will%20result%20in-,kube%2Dcontroller%2Dmanager,-being%20run%20with%20the) in charm config options at charmhub.io.

<img width="705" height="270" alt="Image" src="https://github.com/user-attachments/assets/ee05cb5d-d467-4124-bfea-d512dc00dd3e" />

[This line](https://github.com/canonical/k8s-operator/blob/main/charms/worker/k8s/charmcraft.yaml#L267) has the error and needs to be fixed.

PR #763, opened 2025-11-25
docs: Fix typo in the charmcraft.yaml

nhennigan:
### Overview

Addresses issue #762 that reported the datastore-extra-args docs string referred to the kube-controller-manager by accident

### Rationale

Our docs need to be accurate


addyess (approved):
LGTM

PR #837, opened 2026-01-08
chore: remove `grafana-agent` from COS test suite

mateoflorido:
## Overview
Test only the `opentelemetry-collector:2/stable` metrics agent.

## Rationale

Following the deprecation of `grafana-agent`, we should stop testing both metrics agent and focus only on `opentelemetry-collector:2/stable`.

### DIFF
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -1,7 +1,7 @@
 header:
   - license:
       copyright-owner: Canonical Ltd.
-      copyright-year: 2025
+      copyright-year: 2026
       content: |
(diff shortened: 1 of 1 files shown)


mateoflorido:
> We'll probably never need to go back to `grafana-agent` 1/stable.
> 
> There may come a day we need to test multiple versions of otel-collector. Should we keep the `METRICS_AGENTS` list?

Okay, let's keep it for now in case we need to support multiple agents again.

addyess (approved):

PR #857, opened 2026-01-20
ci: only raise a PR if both builds match

mateoflorido:
## Overview

Only create a PR when both AMD64 and ARM64 have new revisions and their commit digests match. Also, change the commit title to follow the conventional commits format.

### Rationale

Currently, we are getting PRs that only target one architecture, creating deltas in a charm revision between the two supported architectures. This adds to our maintenance load and waste cycles in our self-hosted runners.


addyess (approved):
I really think this is the right auto-update scenario.  Thanks @mateoflorido

PR #913, opened 2026-04-07
ci: pin GitHub Actions to commit SHAs

louiseschmidtgen:
## Summary

Pin all GitHub Actions to full commit SHAs for supply-chain security hardening, replacing mutable tag references.

## Why

Mutable tags (e.g. `@v4`) can be moved by upstream maintainers or by an attacker who compromises an action repo. Pinning to a SHA ensures workflows always run exactly the reviewed code.

This follows GitHub's own security hardening guidance and improves our OpenSSF Scorecard rating.\n\n

## Changes

Pinned all third-party and GitHub-maintained actions in workflow files to full 40-character commit SHAs- Pinned reusable workflow references (operator-workflows, k8s-workflows, charming-actions) to commit SHAs - Preserved original tag/branch versions in inline comments for readability (e.g. ` v4`, ` main`) - Left already-pinned actions unchanged

## Testing

Verified all `uses:` references in `.github/workflows/*.yaml` now use immutable SHAs

### DIFF
--- a/.github/workflows/auto-merge-prs.yaml
+++ b/.github/workflows/auto-merge-prs.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   auto-merge-successful-prs:
     name: Auto-merge K8s charm PRs
-    uses: canonical/k8s-workflows/.github/workflows/auto-merge-successful-prs.yaml@main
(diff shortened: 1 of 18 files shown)


bschimke95 (approved):
All GitHub Action SHAs verified against their claimed version tags. Changes are correct and consistent.

PR #952, opened 2026-06-18
feat: add kube-proxy-enabled config option

HomayoonAlimohammadi:
### Overview

This PR adds a `kube-proxy-enabled` config option as a knob for admins to turn on/off kube-proxy (e.g. in case of a kube-proxy replacement solution like Cilium).

### DIFF
--- a/charms/worker/k8s/charmcraft.yaml
+++ b/charms/worker/k8s/charmcraft.yaml
@@ -402,6 +402,19 @@ config:
         Space-separated list of namespaces to check for unready pods. This
         is used to determine the status of the cluster and its components.
         If not set, no namespaces will be checked.
+    kube-proxy-enabled:
+      type: string
+      default: "auto"
+      description: |
+        Determines if kube-proxy is going to be running on the nodes.
+        The admins can disable kube-proxy in case of a replacement like Cilium.
+
+        Allowed values are "auto", "false", and "true".
+
+        * auto:     The cluster automatically decides whether or not kube-proxy should be enabled.
+        * false:    Disable kube-proxy in the cluster nodes.
+        * true:     Enable kube-proxy in the cluster nodes.
+
 
 resources:
   snap-installation:
--- a/charms/worker/k8s/src/config/cluster.py
+++ b/charms/worker/k8s/src/config/cluster.py
@@ -5,6 +5,7 @@
 
 """Cluster configuration options."""
 
+import logging
 from typing import Optional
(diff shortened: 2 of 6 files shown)


charms/worker/k8s/charmcraft.yaml
```diff
@@ -402,6 +402,12 @@ config:
         Space-separated list of namespaces to check for unready pods. This
         is used to determine the status of the cluster and its components.
         If not set, no namespaces will be checked.
+    kube-proxy-enabled:
+      type: boolean
+      default: true
```
ktsakalozos-canonical: This should not be "true" by default. Can we have 3 values:

* auto: whatever the cluster thinks we should have here
* true: enable kube-proxy
* false: disable kube-proxy

HomayoonAlimohammadi: good point. I'll take care of it

charms/worker/k8s/src/config/cluster.py
```diff
@@ -80,6 +83,17 @@ def _assemble_network(charm: ops.CharmBase, assembled: UserFacingClusterConfig):
         network = assembled.network = NetworkConfig()
     network.enabled = literals.NETWORK_ENABLED.get(charm)
 
+    kube_proxy_enabled = literals.KUBE_PROXY_ENABLED.get(charm).lower()
+    if kube_proxy_enabled not in ["auto", "false", "true"]:
```
canonicalmateo: Could you please add the list to the literals module?

charms/worker/k8s/src/config/cluster.py
```diff
@@ -80,6 +83,17 @@ def _assemble_network(charm: ops.CharmBase, assembled: UserFacingClusterConfig):
...
+        log.error(f"invalid value for kube-proxy-enabled config option: {kube_proxy_enabled}")
+
+    if kube_proxy_enabled == "true":
+        network.kube_proxy_enabled = True
+    elif kube_proxy_enabled == "false":
+        network.kube_proxy_enabled = False
+    # we do not set network.kube_proxy_enabled when the config option is "auto"
+    # so that the cluster decides automatically.
```
canonicalmateo: Similarly here, you can use the literals  here and add more logging to surface what's happening to the administrator.

ktsakalozos (approved):

ktsakalozos-canonical (approved):

PR #976, opened 2026-08-10
feat: migrate from `rocks.c.c` to `ghcr.io`

canonicalmateo:
Applicable spec: <link>

### Overview

Repoint the registry to `rocks.c.c` to `ghcr.io`.

### Rationale

We are deprecating rocks.c.c, so we need to repoint all the images to the new registry.


louiseschmidtgen (approved):
lgtm

PR #967, opened 2026-07-16
feat: support YAML cluster-annotations for BGP peers

louiseschmidtgen:
## Summary
- parse `cluster-annotations` as a YAML mapping first
- keep backward compatibility by falling back to legacy space-separated `key=value` parsing
- wire annotations into assembled cluster config so they are sent to k8sd
- document YAML format (preferred) and legacy format (supported)
- add unit tests for YAML, multiline BGP, legacy format, and invalid legacy handling

## How to set BGP peers via charm config
This branch supports setting BGP peers through `cluster-annotations` YAML.

Example used in manual testing:

```bash
juju config k8s \
  load-balancer-enabled=true \
  load-balancer-bgp-mode=true \
  load-balancer-bgp-local-asn=65000 \
  cluster-annotations="$(cat <<'EOT'
k8sd/v1alpha1/metallb/bgp-peers: |
  - peerAddress: 192.0.2.1
    peerASN: 65001
    myASN: 65000
    nodeSelector:
      topology.kubernetes.io/zone: zone-a
k8sd/v1alpha1/metallb/advertise-all-pools: \"true\"
EOT
)"
```

## Backward compatibility
Legacy format remains supported for simple annotations:

```bash
juju config k8s cluster-annotations='key1=value1 key2=value2'
```

## Verification commands
```bash
juju config k8s cluster-annotations --format yaml
juju config k8s load-balancer-enabled --format yaml
juju config k8s load-balancer-bgp-mode --format yaml
juju config k8s load-balancer-bgp-local-asn --format yaml
```

## Testing
unit tests and 
<details>
<summary> manual testing (expand to see BGPPeer and BGPAdvertisement CRDs) </summary>

```
juju exec --unit k8s/1 -- 'sudo k8s kubectl get BGPPeer -A'
juju exec --unit k8s/1 -- 'sudo k8s kubectl get BGPPeer -A -o yaml'
juju exec --unit k8s/1 -- 'sudo k8s kubectl get BGPAdvertisement -A'
juju exec --unit k8s/1 -- 'sudo k8s kubectl get BGPAdvertisement -A -o yaml'
NAMESPACE        NAME                                     ADDRESS         ASN     BFD PROFILE   MULTI HOPS
metallb-system   metallb-loadbalancer-ck-loadbalancer     10.116.3.164    65001                 
metallb-system   metallb-loadbalancer-ck-loadbalancer-1   10.116.37.164   65002                 
[...]

PR #508, opened 2025-06-13
Prevent k8s snap restarts if the cluster config doesn't change

addyess:
Applicable spec: <link>

### Overview

Prevents the k8s snap from restarting when the charm reconciles to the same cluster-config state.

### Rationale

The cilium pods do quite a dance to restart every time the reconcile hook runs

### Juju Events Changes

None

### Module Changes

creation of `config.cluster` and `config.options` to separate more logic out of charm.py

### Library Changes

None

### Checklist

- [ ] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [ ] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [ ] The documentation is generated using `src-docs`
- [ ] The documentation for charmhub is updated
- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`)

PR #977, opened 2026-08-12
fix: restore cluster-annotations wiring in assemble_cluster_config

louiseschmidtgen:
## What

Restores the `cluster-annotations` charm config option, which currently has **zero effect** on the cluster on `main`.

## Why

This is a regression from #508 ("Prevent k8s snap restarts if the cluster config doesn't change"), which split `_assemble_cluster_config()` out of `charm.py` into a new `config/cluster.py` module. Every other config field (`dns`, `gateway`, `ingress`, `load_balancer`, `metrics_server`, `network`, `local_storage`) got its own `_assemble_*()` step ported into the new module — annotations did not. `_get_valid_annotations()` was left behind in `charm.py`, orphaned with no caller.

Before #508, this config option worked (introduced in #154, survived the #169 refactor). There was never any test coverage for it, so the regression went unnoticed.

## What this PR does (and doesn't do)

This is the **minimal fix**: adds `_assemble_annotations()` to `config/cluster.py` following the same pattern as the other `_assemble_*()` helpers, wired into `assemble_cluster_config()`, reusing the existing `_get_valid_annotations()` parser. Restores parity with the pre-#508 behavior — **legacy space-separated `key=value` format only**.

It does **not** add YAML/nested-annotation support (needed for multi-value keys like BGP peers) — that's a separate, larger feature discussed in #967, which this PR does not attempt to replace or preclude.

## Testing

- Added `test_configure_annotations` (verifies `cluster-annotations` config is now assembled into `UserFacingClusterConfig.annotations`)
- Added `test_configure_annotations_not_overwritten_when_empty` (verifies existing annotations, e.g. set out-of-band via the k8s CLI, are preserved when the charm config is unset — matches k8sd's key-level annotation merge semantics)
- `tox -e unit -- tests/unit/config/test_cluster.py` — all 8 tests pass (4 pre-existing + 2 new x 2 param variants)
- `tox -e lint` — clean (ruff + mypy)

### DIFF
--- a/tests/integration/test_k8s.py
+++ b/tests/integration/test_k8s.py
@@ -139,8 +139,21 @@ def test_cluster_annotations(
         f"Expected annotation {testname}=true to reach k8sd, got: {annotations}"
     )
 
+    # Remove the annotation via the key=- deletion marker. Resetting the charm
+    # config to its default (empty) value alone would leave the annotation
+    # stored in the cluster config.
+    removal_config = {"cluster-annotations": f"{testname}=-"}
     with fast_forward(k8s_cluster, ONE_MIN):
-        k8s_cluster.config("k8s", reset=list(annotation_config))
+        k8s_cluster.config("k8s", removal_config)
+        wait_active(k8s_cluster, "k8s", timeout=timeout * 60)
+
+    annotations = get_k8sd_cluster_config_annotations(k8s_cluster, leader)
+    assert testname not in annotations, (
+        f"Expected annotation {testname} to be removed from k8sd, got: {annotations}"
+    )
+
+    with fast_forward(k8s_cluster, ONE_MIN):
+        k8s_cluster.config("k8s", reset=list(removal_config))
         wait_active(k8s_cluster, "k8s", timeout=timeout * 60)
 
 


charms/worker/k8s/src/config/cluster.py
```diff
@@ -127,3 +131,16 @@ def _assemble_load_balancer(charm: ops.CharmBase, assembled: UserFacingClusterCo
...
+    """Populate cluster annotations from the cluster-annotations charm config.
+
+    Annotations are only overwritten when the cluster-annotations charm config
+    is non-empty. When it is unset, existing cluster annotations (e.g. those
+    set out-of-band via the k8s CLI) are left untouched.
+    """
+    annotations = charm._get_valid_annotations()
+    if annotations is None:
+        return
+    assembled.annotations = annotations
```
canonicalmateo: k8sd merges new annotation with the existing ones. I think the charm is comparing only its configured annotations against the full stored map, so unrelated annotations make every reconciliation appear changed and will trigger multiple PUTs against the API.

tests/unit/config/test_cluster.py
canonicalmateo, line 1: We need to align with the implementation in k8sd:

https://github.com/canonical/k8sd/blob/4071ef5aa87970254d669b1a05a037005015349b/pkg/k8sd/types/cluster_config_merge_util.go#L42-L67

The mechanism is that the annotations are patching only the existing resource. If we want to remove them, then the value should be `key=-`. In this case, we should also improve the docs to surface the behavior.

canonicalmateo: Could you please add an idempotency test? We should test that the cluster containing both oob and charm managed annotations are handled correctly if we perform multiple reconciliations.

charms/worker/k8s/src/config/cluster.py
```diff
@@ -21,6 +21,9 @@
     UserFacingClusterConfig,
 )
 
+if TYPE_CHECKING:  # pragma: no cover
+    from charm import K8sCharm
+
```
canonicalmateo: This is coupling the module to the concrete charm, maybe you can pass the annotations as a parameter to avoid using the private method (i.e., `_assemble_annotations`).

tests/unit/config/test_cluster.py
canonicalmateo: Also, do we have a test to validate malformed input and the behavior in the charm?

canonicalmateo (changes requested):
Thanks for the PR, Louise! Nice catch. I've left some comments and suggestions.

charms/worker/k8s/src/config/cluster.py
louiseschmidtgen: Fixed in ba95807. `_assemble_annotations` now merges the charm-managed annotations onto the existing cluster annotations (mirroring k8sd's `mergeAnnotationsField`) instead of replacing them. Once k8sd stores the merged map, the assembled config equals the stored one, so reconciliations no longer report perpetual changes or trigger repeated PUTs. Covered by `test_configure_annotations_merged_with_out_of_band` and `test_configure_annotations_idempotent`.

charms/worker/k8s/src/config/cluster.py
louiseschmidtgen: Done in ba95807. `assemble_cluster_config()` now takes the parsed annotations as a parameter and `_assemble_annotations()` no longer touches the charm — the `TYPE_CHECKING` import of `K8sCharm` is removed.

tests/unit/config/test_cluster.py
louiseschmidtgen: Aligned with k8sd in ba95807: annotations now patch the existing map, and `key=-` removes an annotation. The deletion marker is forwarded to k8sd only while the key is still present in the stored config, so repeated reconciliations converge. The merge behavior and the `key=-` removal syntax are now documented in the `cluster-annotations` description in charmcraft.yaml.

louiseschmidtgen: Added `test_configure_annotations_malformed` in ba95807 (parametrized: `malformed`, `key1=`, `=value1`), asserting `ReconcilerError` and `BlockedStatus("Invalid Annotations")`. Writing this test surfaced a real bug: a pair without `=` raised a raw `ValueError` from the tuple unpack, escaping the charm's `@on_error` handling. The parser now uses `str.partition()` so all malformed input raises `ReconcilerError`.

louiseschmidtgen: Added `test_configure_annotations_idempotent` in ba95807: a cluster holding both out-of-band and charm-managed annotations converges after one reconciliation, and subsequent reconciliations detect no config change. `test_configure_annotations_removal` covers the same property for the `key=-` deletion path.

tests/integration/test_k8s.py
```diff
@@ -120,6 +121,29 @@ def test_nodes_labelled(request: pytest.FixtureRequest, k8s_cluster: jubilant.Ju
...
+        wait_active(k8s_cluster, "k8s", timeout=timeout * 60)
+
+    annotations = get_k8sd_cluster_config_annotations(k8s_cluster, leader)
+    assert annotations.get(testname) == "true", (
+        f"Expected annotation {testname}=true to reach k8sd, got: {annotations}"
+    )
+
+    with fast_forward(k8s_cluster, ONE_MIN):
+        k8s_cluster.config("k8s", reset=list(annotation_config))
+        wait_active(k8s_cluster, "k8s", timeout=timeout * 60)
```
canonicalmateo, line 157: One last thing, can we add another clause to remove the annotation and verify it's gone via the socket? This is going to reset the value to its default (i.e., empty), but the annotation will persist in the cluster config.

canonicalmateo (commented):
Great! Just a minor fix before we merge it

tests/integration/test_k8s.py
louiseschmidtgen: Done in b1ddd9b. `test_cluster_annotations` now sets the `key=-` deletion marker, waits for active, and asserts via the k8sd control socket that the annotation is gone from the stored cluster config. The final reset then restores the charm config default — as you noted, resetting to empty alone would have left the annotation persisted.

canonicalmateo (approved):
LGTM! Thanks Louise

PR #982, opened 2026-08-13
fix: use git-ls for snaps autoupdate workflow

canonicalmateo:
## Overview

Replaced the use of the `octokit/request-action` for listing branches with a direct `git ls-remote` command, reducing dependencies and simplifying the workflow.


zhey16 (approved):
lgtm

PR #1012, opened 2026-09-16
feat(terraform)!: Support juju provider v1 and v2

hue-canonical:
Juju provider v1.0.0 replaced model names with model UUIDs on model-scoped resources and removed the `model` attribute from juju_application. The `< 1.0.0` constraint kept consumers such as k8s-bundles pinned to provider 0.x.

Rename the `model` input of the k8s and k8s-worker modules to `model_uuid` and allow provider versions >= 1.0.0, < 3.0.0. Consumers must now pass the model UUID, e.g. juju_model.<name>.uuid.

Require provider >= 2.3.1 in the integration test root module, which consumes k8s-bundles, and import existing models by UUID as provider v1+ expects.


tests/integration/terraform/setup.py
```diff
@@ -210,12 +213,13 @@ def main() -> None:
     print("Initializing Terraform...")
     tf_run(args.terraform_module_path, ["init", "--upgrade"])
 
-    # Import Existing Model if it exists
-    if model_exists(args.model):
+    # Import Existing Model if it exists; juju provider v1+ imports models by UUID
+    uuid = model_uuid(args.model)
```
amc94, line 217: you don't need to import the model, you can make `model_uuid` a variable and conditionally use an external one vs creating one.

hue-canonical: based on our discussion in MM, the changes are fine as is b/c they're just singular modules.

tests/integration/test_upgrade.py
```diff
@@ -22,7 +22,7 @@
 # NOTE: Must satisfy the k8s_service "upgrade_supported" range in
 # charms/worker/k8s/src/literals.py -- the charm refuses to upgrade from outside it,
 # leaving the control-plane leader blocked. Bump this whenever that range moves.
-CHARM_UPGRADE_FROM = os.environ.get("JUJU_DEPLOY_CHANNEL", "1.33/beta")
+CHARM_UPGRADE_FROM = os.environ.get("JUJU_DEPLOY_CHANNEL", "1.35/beta")
```
louiseschmidtgen: What does this change?
Is this supposed to test the full upgrade story from 1.32 or where is this used?

louiseschmidtgen (approved):

tests/integration/test_upgrade.py
hue-canonical: This was a wrong fix that I made and I reverted it back. Basically we need to set it to 1.35 when 1.36 is out (it should've been 1.34, but we've missed that)

PR #1013, opened 2026-09-16
feat(terraform)!: [1.36] Support juju provider v1 and v2

hue-canonical:
Juju provider v1.0.0 replaced model names with model UUIDs on model-scoped resources and removed the `model` attribute from juju_application. The `< 1.0.0` constraint kept consumers such as k8s-bundles pinned to provider 0.x.

Rename the `model` input of the k8s and k8s-worker modules to `model_uuid` and allow provider versions >= 1.0.0, < 3.0.0. Consumers must now pass the model UUID, e.g. juju_model.<name>.uuid.

Require provider >= 2.3.1 in the integration test root module, which consumes k8s-bundles, and import existing models by UUID as provider v1+ expects.


amc94 (approved):
+1

louiseschmidtgen (approved):

